PR #8, opened 2023-10-09
Namespace Management APIs

anekkanti:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // environment of the namespace. [optional]
+    // NOTE: currently there is no additional SLA or functional guarantee implied by the value of this field.
+    // supported environments: dev, test, prod
+    string environment = 6;
+    // codec server property spec needed for user to set and retrieve [optional]
+    CodecServerPropertySpec codec_spec = 7;
+    // the regions where the namespace is (or will be) located [optional]
+    repeated string passive_regions = 8;
+    // the export sink specifications keyed on the sink name [optional]
+    map<string, temporal.api.cloud.sink.v1.ExportSinkSpec> export_sinks = 9;
```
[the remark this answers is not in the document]
robzienert: For my particular use case, it would be nice to be able to reference a common export sink config since we use the same config globally across all namespaces (as history export is a platform-managed feature here).

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // the namespace specification
+    NamespaceSpec spec = 3;
+    // the current state of the namespace
+    string state = 4;
+    // The id of the async operation that is creating/updating/deleting the namespace, if any
+    string async_operation_id = 5;
+    // the web uri for the namespace
+    NamespaceURI uri = 6;
+    // the envelope is a list of service level agreements (SLAs) that can be provided around a given namespace
+    NamespaceEnvelope envelope = 7;
```
robzienert, line 86: Awesome.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -107,3 +108,94 @@ message GetAsyncOperationResponse {
...
+    // the prefix to use for the namespace
+    // will create a namespace that's available at '<namespace_prefix>.<account>.tmprl.cloud:7233'
+    string namespace_prefix = 1;
+    // the namespace specification
+    temporal.api.cloud.namespace.v1.NamespaceSpec spec = 2;
+    // The id to use for this async operation - optional
+    string async_operation_id = 3;
+}
+
+message CreateNamespaceResponse {
```
anekkanti, line 122: Add `namespace`

temporal/api/cloud/sink/v1/message.proto
```diff
@@ -0,0 +1,45 @@
...
+    string health = 2;
+    // The error message of the sink if any
+    string error_message = 3;
+    // The latest data export time
+    google.protobuf.Timestamp latest_data_export_time = 4;
+    // The latest health check time
+    google.protobuf.Timestamp last_health_check_time = 5;
+}
+
+message S3Spec{
```
cretz, line 32:
```suggestion
message S3Spec {
```

temporal/api/cloud/sink/v1/message.proto
```diff
@@ -0,0 +1,45 @@
+syntax = "proto3";
+
+package temporal.api.cloud.sink.v1;
+
+option go_package = "go.temporal.io/api/cloud/sink/v1;sink";
+
+import "google/protobuf/timestamp.proto";
+
+message ExportSinkSpec {
```
cretz, line 9: Consider a unique, user-provided identifier for a sink (e.g. a name) if you want to support updating of these inside a list. Not required of course, but can be easier on you, the UI, and the user.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // Environment of the namespace. - optional
+    // NOTE: currently there is no additional SLA or functional guarantee implied by the value of this field.
+    // supported environments: dev, test, prod
+    string environment = 6;
+    // Codec server property spec needed for user to set and retrieve - optional
+    CodecServerPropertySpec codec_spec = 7;
+    // The regions where the namespace is (or will be) located - optional
+    repeated string passive_regions = 8;
+    // The export sink specifications keyed on the sink name - optional
+    map<string, temporal.api.cloud.sink.v1.ExportSinkSpec> export_sinks = 9;
```
cretz, line 54: Arguably "export" is too generic of a word for this. Is this just history export? Or is this also log export, audit export, user export, and all other exports we may support henceforth?

temporal/api/cloud/namespace/v1/message.proto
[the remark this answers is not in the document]
anekkanti: @robzienert thanks for the this feedback. 
We are tracking this as a feature as part of the "hierarchy" story where various configurations like ca-certs, search-attrs, export configs can trickle down a hierarchy sub-tree as defaults...
cc: @jlacefie @sergeybykov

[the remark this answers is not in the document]
jlacefie: TY - Rob.  We released the first Export per Namespace with the understanding that some Namespace owners would like isolation at the S3 bucket. In a subsequent release we can provide Account level affordances for a default.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,89 @@
...
+    // The name of the attribute is the key and the type is the value
+    // Supported attribute types: text, keyword, int, double, bool, datetime, keyword_list
+    map<string, string> search_attributes = 4;
+    // Certificate filters which, if specified, only allow connections from client certificates
+    // whose distinguished name properties match at least one of the filters
+    repeated CertificateFilterSpec certificate_filters = 5;
+    // Environment of the namespace. - optional
+    // NOTE: currently there is no additional SLA or functional guarantee implied by the value of this field.
+    // supported environments: dev, test, prod
+    string environment = 6;
```
jlacefie, line 46: May we not include environment yet please?  This seems premature and something that can be added later after we are aligned with the way we will expose hierarchy containers and metadata to users.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,89 @@
+syntax = "proto3";
+
+package temporal.api.cloud.namespace.v1;
+
+option go_package = "go.temporal.io/api/cloud/namespace/v1;namespace";
+
+import "google/protobuf/timestamp.proto";
+
+message CertificateFilterSpec {
```
jlacefie, line 9: pls add comment describing the use of certificate filters

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    string common_name = 1;
+    // The organization in the certificate
+    string organization = 2;
+    // The organizational_unit in the certificate
+    string organizational_unit = 3;
+    // The subject_alternative_name in the certificate
+    string subject_alternative_name = 4;
+}
+
+message CodecServerPropertySpec {
```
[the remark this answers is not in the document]
jlacefie: +1 for removing property
The Cloud UI referes to this field as Codec Server only. 
Recommend adding a comment to clarify the CodecServer purpose like in the UI 

Enter a codec server endpoing to decode payloads for all users interacting with this Namespace in the UI.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,89 @@
...
+    // Server endpoints
```
jlacefie, line 21: nit - codec server endpoint

Also, does this take in 1 endpoint or many? the plural endpoints in the comment is ambiguous

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,89 @@
...
+    string endpoint = 1;
+    // Whether to pass access token, i.e. jwt
```
jlacefie, line 23: nit - comment should align to the UI to indicate where the JWT ends up. 
UI comment - Pass the user access token with your endpoint.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    bool pass_access_token = 2;
+    // Whether to include credentials
+    bool include_credentials = 3;
```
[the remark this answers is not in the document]
jlacefie: +1 ^ CORS should be explicit.  Pls update the comment as well.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,89 @@
...
+}
+
+message NamespaceSpec {
+    // The region where the namespace is (or will be) active
+    string region = 1;
+    // The base64 encoded ca cert(s) that the clients can use for authentication and authorization
+    string accepted_client_ca = 2;
+    // The number of days the workflows data will be retained for
```
jlacefie, line 34: comment nit - align with the UI comment to indicate the impact of changing retention period
UI description
_Changes to the retention period may impact your storage costs.
Any changes to the retention period will be applied to all new running workflows._

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,89 @@
...
+    int32 retention_days = 3;
```
jlacefie, line 35: align to the UI - retention_period could add days in the field name as well if desired

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,89 @@
...
+message CodecServerPropertySpec {
+    // Server endpoints
+    string endpoint = 1;
+    // Whether to pass access token, i.e. jwt
+    bool pass_access_token = 2;
+    // Whether to include credentials
+    bool include_credentials = 3;
+}
+
+message NamespaceSpec {
```
jlacefie, line 29: please include cost_center as we are adding this with the cost center per namespace initiative now
type string
optional

comment - optionally tag a namespace with a cost center. Adding a cost center to a namespace allows for cost center aggregations to appear on an invoice and in usage metrics. This is useful if you want to attribute Namespace usage and costs to a specific cost center.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // whose distinguished name properties match at least one of the filters
+    repeated CertificateFilterSpec certificate_filters = 5;
+    // Environment of the namespace. - optional
+    // NOTE: currently there is no additional SLA or functional guarantee implied by the value of this field.
+    // supported environments: dev, test, prod
+    string environment = 6;
+    // Codec server property spec needed for user to set and retrieve - optional
+    CodecServerPropertySpec codec_spec = 7;
+    // The regions where the namespace is (or will be) located - optional
+    repeated string passive_regions = 8;
```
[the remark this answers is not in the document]
jlacefie: is this for Global Namespace?  If so, is passive_region the right term here?  This feels like a one-way door. 

Also, will we offer multiple regions in the future? If so, passive seems restricting as it implies one option. 
This item in particular needs more thought IMO.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,89 @@
...
+}
+
+message NamespaceURI {
```
jlacefie, line 53: what is the purpose of this message?

temporal/api/cloud/namespace/v1/message.proto
jlacefie: Please include Export configuration as a part of this spec or in the Namespace message as a separate message type

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // The current state of the namespace
+    string state = 4;
+    // The id of the async operation that is creating/updating/deleting the namespace, if any
+    string async_operation_id = 5;
+    // The web uri for the namespace
+    NamespaceURI uri = 6;
+    // The envelope is a list of service level agreements (SLAs) that can be provided around a given namespace
+    NamespaceEnvelope envelope = 7;
+    // Allowed principals is a list of principals that allowed to access the private links on the namespace
+    repeated string allowed_principals = 8;
```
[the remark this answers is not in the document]
jlacefie: It looks like main includes Identity. Could we use Identity here if that's the intent of this field, please?

temporal/api/cloud/namespace/v1/message.proto
jlacefie: @robzienert TY for the comment here.  What is exciting about this one for you? Also, do you have an opinion on the term Envelope in this context?

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+message Namespace {
+    // The namespace name
+    string namespace = 1;
+    // The current version of the namespace specification
+    // The next update operation will have to include this version
+    string resource_version = 2;
+    // The namespace specification
+    NamespaceSpec spec = 3;
+    // The current state of the namespace
+    string state = 4;
```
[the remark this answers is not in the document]
jlacefie: @anekkanti is there a reason for not using enums as described here ^ is that an explicit decision?

temporal/api/cloud/sink/v1/message.proto
jlacefie, line 9: Where do users apply the sink message to a Namespace?  It doesn't appear to be used in either the service or namespace protos?  Please make sink an option when creating a namespace and provide users the ability to add a sink to a namespace at a later time.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -107,3 +108,95 @@ message GetAsyncOperationResponse {
     // The async operation
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
+
+message CreateNamespaceRequest {
+    // The prefix to use for the namespace
+    // will create a namespace that's available at '<namespace_prefix>.<account>.tmprl.cloud:7233'
+    string namespace_prefix = 1;
```
jlacefie, line 115: why is the namespace_prefix not a part of the namespace spec?

temporal/api/cloud/namespace/v1/message.proto
anekkanti: @jlacefie there is a comment in the namespace spec object where this struct is used.

temporal/api/cloud/namespace/v1/message.proto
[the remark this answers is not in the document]
anekkanti: Changed to `CodecServerSpec` as suggested.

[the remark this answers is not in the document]
anekkanti: Updated comment. Please review.

temporal/api/cloud/namespace/v1/message.proto
anekkanti: Its just one, fixed the comments.

temporal/api/cloud/namespace/v1/message.proto
anekkanti: Ack fixed, as mentioned in the UI.

temporal/api/cloud/namespace/v1/message.proto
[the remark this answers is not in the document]
anekkanti: fixed.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+message CodecServerPropertySpec {
+    // Server endpoints
+    string endpoint = 1;
+    // Whether to pass access token, i.e. jwt
+    bool pass_access_token = 2;
+    // Whether to include credentials
+    bool include_credentials = 3;
+}
+
+message NamespaceSpec {
```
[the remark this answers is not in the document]
anekkanti: Marked each field if they are optional or immutable.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // The region where the namespace is (or will be) active
+    string region = 1;
+    // The base64 encoded ca cert(s) that the clients can use for authentication and authorization
+    string accepted_client_ca = 2;
```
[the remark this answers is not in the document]
anekkanti: Ack, like the restructuring. Fixed.

temporal/api/cloud/namespace/v1/message.proto
anekkanti: ack fixed as suggested.

temporal/api/cloud/namespace/v1/message.proto
anekkanti: ack fixed.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // The number of days the workflows data will be retained for
+    int32 retention_days = 3;
+    // The custom search attributes to use for the namespace
+    // The name of the attribute is the key and the type is the value
+    // Supported attribute types: text, keyword, int, double, bool, datetime, keyword_list
+    map<string, string> search_attributes = 4;
```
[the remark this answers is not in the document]
anekkanti: Yes they will be mapped to the `temporal.api.enums.v1.IndexedValueType`

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // Certificate filters which, if specified, only allow connections from client certificates
+    // whose distinguished name properties match at least one of the filters
+    repeated CertificateFilterSpec certificate_filters = 5;
+    // Environment of the namespace. - optional
+    // NOTE: currently there is no additional SLA or functional guarantee implied by the value of this field.
+    // supported environments: dev, test, prod
+    string environment = 6;
+    // Codec server property spec needed for user to set and retrieve - optional
+    CodecServerPropertySpec codec_spec = 7;
```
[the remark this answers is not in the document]
anekkanti: Changed to `coded_server`.

temporal/api/cloud/namespace/v1/message.proto
anekkanti: Its the information that the customer will needs to locate the namespace's UI and the URL that the temporal client should connect to.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // The regions where the namespace is (or will be) located - optional
+    repeated string passive_regions = 8;
+    // The export sink specifications keyed on the sink name - optional
+    map<string, temporal.api.cloud.sink.v1.ExportSinkSpec> export_sinks = 9;
+}
+
+message NamespaceURI {
```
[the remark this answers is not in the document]
anekkanti: renamed to `NamespaceEndpoints`

temporal/api/cloud/namespace/v1/message.proto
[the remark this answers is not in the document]
anekkanti: This is an explicit decision that we took based on our experience with the existing internal APIs.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    // The namespace specification
+    NamespaceSpec spec = 3;
+    // The current state of the namespace
+    string state = 4;
+    // The id of the async operation that is creating/updating/deleting the namespace, if any
+    string async_operation_id = 5;
+    // The web uri for the namespace
+    NamespaceURI uri = 6;
+    // The envelope is a list of service level agreements (SLAs) that can be provided around a given namespace
+    NamespaceEnvelope envelope = 7;
```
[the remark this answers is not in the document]
anekkanti: Why? 
We do talk about envelops when talking about our cloud offering: 
https://docs.temporal.io/cloud/operating-envelope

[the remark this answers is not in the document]
cretz: I think it probably wasn't necessarily a good choice of words there either (in code, an envelope is often a "wrapper" or container something is in when making calls), but if this is general SaaS nomenclature for limits, ok. I was just confused as a dev. Maybe at least make it `OperatingEnvelope`.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -6,13 +6,15 @@ option go_package = "go.temporal.io/api/cloud/cloudservice/v1;cloudservice";
...
-    // The requested size of the page to retrieve
+    // The requested size of the page to retrieve - optional.
+    // Cannot exceed 1000. Defaults to 100. 
     int32 page_size = 1;
-    // The page token
+    // The page token if this is continuing from another response - optional.
     string page_token = 2;
-    // Optional field to filter users by email address
+    // Filter users by email address - optional
     string email = 3;
```
tminusplus, line 18: Completely unrelated, feel free to punt this comment for another day.

Should we have allowed a list of emails? Then this can operate as a batch Get operation (as opposed to just single email). Ditto for IDs.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -107,3 +109,99 @@ message GetAsyncOperationResponse {
...
+}
+
+message GetNamespacesRequest {
+    // The requested size of the page to retrieve.
+    // Cannot exceed 1000. 
+    // Optional, defaults to 100. 
+    int32 page_size = 1;
+    // The page token if this is continuing from another response.
+    // Optional, defaults to empty.
+    string page_token = 2;
```
tminusplus, line 135: Will we add more filters in a follow-up? Things like region (or regions) etc.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+message AWSPrivateLink {
+    // The list of principals that are allowed to access the private links on the namespace.
+    repeated string allowed_principal_arns = 1;
+    // The list of private links, if any, that are associated with the namespace.
+    repeated string vpc_endpoint_service_names = 2;
+}
+
+message Namespace {
+    // The namespace identifier.
+    string namespace = 1;
```
tminusplus, line 98: I wonder if calling `namespace` in all these requests as `id` is clearer. Just because of the confusion around `namespace name` vs `<name>.<account_id>`.

That said, I understand that this aligns with what we call it in the OSS APIs.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+    // The current state of the namespace.
+    string state = 4;
+    // The id of the async operation that is creating/updating/deleting the namespace, if any.
+    string async_operation_id = 5;
+    // The endpoints for the namespace.
+    Endpoints endpoints = 6;
+    // The envelope is a list of service level agreements (SLAs) that can be provided around a given namespace.
+    OperatingEnvelope operating_envelope = 7;
+    // The private link for the namespace, if any.
+    oneof private_link {
```
tminusplus, line 113: In the future, its possible that we have a global namespace across clouds which uses both AWS private link and GCP private service connect (as an example). Or wireguard etc.

We may want to drop the `oneof` to ensure we aren't limited in the protocol.

temporal/api/cloud/cloudservice/v1/request_response.proto
cretz: :+1: At the very least this should clarify whether it's exact match or partial match

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+
+package temporal.api.cloud.namespace.v1;
+
+option go_package = "go.temporal.io/api/cloud/namespace/v1;namespace";
+
+import "temporal/api/cloud/sink/v1/message.proto";
+
+import "google/protobuf/timestamp.proto";
+
+message CertificateFilterSpec {
```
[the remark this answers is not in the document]
cretz: Bump, unsure if this does exact string matches or partial string matches, and unsure whether any field must match or all fields must match.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+message CodecServerSpec {
+    // The codec server endpoint.
+    string endpoint = 1;
+    // Whether to pass the user access token with your endpoint.
+    bool pass_access_token = 2;
+    // Whether to include cross-origin credentials.
+    bool include_cross_origin_credentials = 3;
+}
+
+message GlobalSpec {
```
cretz, line 43:
```suggestion
message GlobalNamespaceSpec {
```

Usually I'd say you don't need the somewhat redundant `Namespace` but you need it here for two reasons: 1) just "global" for options has general meaning (one might think it's some cross-namespace global scope) and 2) the name of the product is "global namespace".

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+    // The base64 encoded ca cert(s) in PEM format that the clients can use for authentication and authorization.
+    // This must only be one value, but the CA can have a chain.
+    string accepted_client_ca = 1;
+    // Certificate filters which, if specified, only allow connections from client certificates whose distinguished name properties match at least one of the filters.
+    // This allows limiting access to specific end-entity certificates.
+    // Optional, default is empty.
+    repeated CertificateFilterSpec certificate_filters = 2;
+}
+
+message CodecServerSpec {
```
cretz, line 34:
```suggestion
message UICodecServerSpec {
```

Mentioned before but now that I think about it, a user may assume this codec server will be used for everything instead of just the UI. Not a strong opinion here though.

temporal/api/cloud/namespace/v1/message.proto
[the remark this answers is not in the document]
cretz: Do I assume if it doesn't say optional then it's required? Arguably also saying something is required is simpler than me having to infer it by a lack of "optional".

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+    MtlsAuthSpec mtls_auth = 4;
+    // The custom search attributes to use for the namespace.
+    // The name of the attribute is the key and the type is the value.
+    // Supported attribute types: text, keyword, int, double, bool, datetime, keyword_list.
+    // NOTE: currently deleting a search attribute is not supported.
+    // Optional, default is empty.
+    map<string, string> custom_search_attributes = 5;
+    // Codec server spec used by UI to decode payloads for all users interacting with this namespace.
+    // Optional, default is unset.
+    CodecServerSpec codec_server = 6;
```
cretz, line 71:
```suggestion
    UICodecServerSpec ui_codec_server = 6;
```

See my comment on the message name itself

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+    // The global namespace spec.
+    // Optional, default is unset.
+    GlobalSpec global = 7;
```
cretz, line 74:
```suggestion
    GlobalNamespaceSpec global_namespace = 7;
```

See my comment on the message name itself

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+}
+
+message Endpoints {
```
cretz, line 77: Would totally support inlining this information in the `Namespace` if you wanted. Not a ton of value separating this out. But no strong opinion.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+
+message Namespace {
+    // The namespace identifier.
+    string namespace = 1;
+    // The current version of the namespace specification.
+    // The next update operation will have to include this version.
+    string resource_version = 2;
+    // The namespace specification.
+    NamespaceSpec spec = 3;
+    // The current state of the namespace.
```
cretz, line 104: If you are not going to use proper proto enumerates, please list all possible enumerates in doc for every one of these stringly-typed enumerates for the benefit of the caller.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+    string state = 4;
+    // The id of the async operation that is creating/updating/deleting the namespace, if any.
+    string async_operation_id = 5;
+    // The endpoints for the namespace.
+    Endpoints endpoints = 6;
+    // The envelope is a list of service level agreements (SLAs) that can be provided around a given namespace.
+    OperatingEnvelope operating_envelope = 7;
```
cretz, line 111: Just want to document here that I somewhat disagree with this seemingly Temporal-specific made up term compared to something like "limits", but no need to change anything

temporal/api/cloud/namespace/v1/message.proto
cretz: While it is a protobuf backwards compatible operation to move a field into a `oneof` later, it's often not a code-gen backwards compatible alteration. I do think it should be decided what compatibility guarantees you want to provide when you do provide them - wire guarantees or code-gen guarantees. If only the former, which is reasonable, then yes can wait on the `oneof`.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+    // The private link for the namespace, if any.
+    oneof private_link {
+    	AWSPrivateLink aws_private_link = 8;
+    }
+    // The date and time when the user was created.
+    google.protobuf.Timestamp created_time = 9;
+    // The date and time when the namespace was last modified.
```
cretz, line 118: Can you clarify whether this will be null or the same as `created_time` if never modified?

temporal/api/cloud/namespace/v1/message.proto
cretz: It's much less clear to try to use `id` because "namespace ID" is another thing. It is good that `namespace` aligns across _all_ Temporal APIs.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -0,0 +1,120 @@
...
+    // The global namespace spec.
+    // Optional, default is unset.
+    GlobalSpec global = 7;
+}
+
+message Endpoints {
+    // The web ui address.
+    string web_address = 1;
+    // The grpc hostport that the temporal workers, clients and tctl connect to.
+    string grpc_hostport = 2;
```
cretz, line 81:
```suggestion
    string grpc_address = 2;
```

My suggestion from the previous GH comment was about altering proto comment, not the field name. Though I have no strong opinion here, but if you do want to alter the field name, `host_port` is two words I think.

temporal/api/cloud/namespace/v1/message.proto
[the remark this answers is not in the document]
jlacefie: nit. indicate the number of regions supported in the comment with the understanding that the comment can change as we open up the possibility of more regions. For context, the plural name indicates more than 1

temporal/api/cloud/namespace/v1/message.proto
jlacefie: +1 - I reached out to the crew-dynamic-namespace-envelop channel and asked if anyone has researched this. IMO, we create minor cognitive friction if we use uncommon terms for common concepts like quotas and limits

jlacefie: asked @sergeybykov for input here as well.  This feedback applies to the message definition on line 84 as well 

recommend using quotas_and_limits

temporal/api/cloud/cloudservice/v1/request_response.proto
anekkanti: Absolutely with time.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -107,3 +108,95 @@ message GetAsyncOperationResponse {
...
+    // The id to use for this async operation - optional
+    string async_operation_id = 4;
+}
+
+message UpdateNamespaceResponse {
+    // The async operation
+    temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
+}
+
+message RenameCustomSearchAttributeRequest {
```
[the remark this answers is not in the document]
anekkanti: After some async conversion we decided to keep it `RenameCustomSearchAttributeRequest`

temporal/api/cloud/namespace/v1/message.proto
anekkanti: I don't know.. its totally possible that non ui client can use the endpoint field... the other fields don't make a lot of sense though.

tminusplus (approved):

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -59,4 +59,63 @@ service CloudService {
...
+
+    // Delete a namespace
+    rpc DeleteNamespace (DeleteNamespaceRequest) returns (DeleteNamespaceResponse) {
+	option (google.api.http) = {
+            delete: "/api/v1/namespaces/{namespace}",
+        };
+    }
+
+    // Get all regions
+    rpc GetRegions (GetRegionsRequest)  returns (GetRegionsResponse) {
```
cretz, line 109:
```suggestion
    rpc GetRegions (GetRegionsRequest) returns (GetRegionsResponse) {
```

Do you have a linter running?

cretz (approved):

cretz:
Approving under the assumption that this is still technically a beta API that can have changes made as we learn more.

PR #10, opened 2023-10-24
aliceyin/add export wf history proto

alice-yin:
- Add export wf history proto
- Add workflow history file

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/buf.yaml
+++ b/buf.yaml
@@ -4,6 +4,10 @@ deps:
 breaking:
   use:
     - FILE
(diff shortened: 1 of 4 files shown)


buf.yaml
alice-yin, line 1: I'm thinking to remove - buf.build/googleapis/googleapis and replace where we're using gooogleapis with what in temporal/api.

PR #11, opened 2023-11-03
Global namespace mgmt

anekkanti:


temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -40,20 +40,17 @@ message CodecServerSpec {
...
-
 message NamespaceSpec {
     // The name to use for the namespace.
     // This will create a namespace that's available at '<name>.<account>.tmprl.cloud:7233'.
     // The name is immutable. Once set, it cannot be changed.
     string name = 1;
-    // The region where the namespace is (or will be) active.
-    // The region is immutable. Once set, it cannot be changed.
-    string region = 2;
+    // The regions where the namespace should be hosted.
```
atihkin, line 48: Regions of availability for the namespace.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -40,20 +40,17 @@ message CodecServerSpec {
...
+    // Specifying more than one region makes the namespace a Global Namespace, which is currently a preview only feature with restricted access.
```
atihkin, line 49: Specifying more than one region makes the namespace "global", which is currently a preview only feature with restricted access.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -40,20 +40,17 @@ message CodecServerSpec {
...
+    // Please reach out to temporal support for more information on Global Namespaces.
```
atihkin, line 50: Please reach out to Temporal support for more information on Global Namespaces.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -40,20 +40,17 @@ message CodecServerSpec {
...
+    // When provisioned the Global Namespace will be active on the first region in the list and passive on the rest.
+    // The regions is immutable. Once set, it cannot be changed.
```
atihkin, line 52: This is true today (in private preview) but regions could be mutable in the future (after namespace creation, user could "edit").

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -40,20 +40,17 @@ message CodecServerSpec {
...
+    repeated string regions = 2;
```
atihkin, line 53: Number of supported regions today is 2. Could be more in the future.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -87,13 +81,20 @@ message Limits {
...
 
-message AWSPrivateLink {
+message AWSPrivateLinkInfo {
     // The list of principals that are allowed to access the private links on the namespace.
     repeated string allowed_principal_arns = 1;
     // The list of private links, if any, that are associated with the namespace.
     repeated string vpc_endpoint_service_names = 2;
 }
 
+message PrivateConnectivity {
```
atihkin, line 91: +1. "Private Connectivity" is a good generic term! It's called PrivateLink in AWS, Private Link in Azure, and Private Service Connect (PSC) in GCP.

temporal/api/cloud/namespace/v1/message.proto
anekkanti: fixed as suggested.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -110,10 +114,9 @@ message Namespace {
     Endpoints endpoints = 6;
     // The limits set on the namespace currently.
     Limits limits = 7;
-    // The private link for the namespace, if any.
-    oneof private_link {
-    	AWSPrivateLink aws_private_link = 8;
-    }
+    // The private connectivity for the namespace by region where they apply, if any.
+    // The region is the key and the private connectivity is the value.
+    map<string, PrivateConnectivity> private_connectivity = 8;
```
cretz, line 119: How is this configured by an API user?

cretz, line 119: Is there only ever going to be one private connection info per region?

cretz, line 119: Is every region-based configuration going to have to be a `map<string, ...`? Maybe region is an object with information about itself.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -40,20 +40,18 @@ message CodecServerSpec {
...
-    // The region where the namespace is (or will be) active.
-    // The region is immutable. Once set, it cannot be changed.
-    string region = 2;
+    // The regions of availability for the namespace.
+    // Specifying more than one region makes the namespace "global", which is currently a preview only feature with restricted access.
+    // Please reach out to Temporal support for more information on global namespaces.
+    // When provisioned the global namespace will be active on the first region in the list and passive on the rest.
+    // Number of supported regions is 2. 
+    // The regions is immutable. Once set, it cannot be changed.
+    repeated string regions = 2;
```
yux0, line 54: do you want to use the RegionID which contains cloud provider and region name?

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -205,3 +206,26 @@ message DeleteNamespaceResponse {
...
+
+message GetRegionRequest {
+    // The id of the region to get.
+    string region = 1;
+}
+
+message GetRegionResponse {
+    // The temporal cloud region.
+    temporal.api.cloud.region.v1.Region region = 1;
+}
```
cretz, line 231: I'm not seeing where these messages are ever used

It might be best, before merging APIs, that the implementation of them also be ready to merge. Often during implementation you find adjustments you need to the APIs. Impl and API should both remain in development while they are in development and then both get merged when they are ready.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -205,3 +206,26 @@ message DeleteNamespaceResponse {
     // The async operation.
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
+
+message GetRegionsRequest {
+    // The requested size of the page to retrieve - optional.
+    // Cannot exceed 1000. Defaults to 100. 
+    int32 page_size = 1;
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
```
cretz, line 215: Where do I get page token from? It's not on the response.

temporal/api/cloud/region/v1/message.proto
cretz, line 1: Does this deserve an entire package?

temporal/api/cloud/namespace/v1/message.proto
anekkanti: added a region id.

temporal/api/cloud/cloudservice/v1/request_response.proto
anekkanti: Added the service entry. 

Absolutely I do not plan on merging these changes into the main branch till I have the implementation ready and deployed on prod.

temporal/api/cloud/cloudservice/v1/request_response.proto
anekkanti: Thanks fixed.

temporal/api/cloud/region/v1/message.proto
anekkanti: A few reasons i can think off:
1) Regions might we something that could potentially be used beyond namespaces for some other kind of resource in the future. 
2) There is a possibility we might need more than what we currently have for a region. Say a customer's private cloud region. 

But since these are not that well defined, i am willing to move it to the namespace package if you feel strongly.

cretz: > Regions might we something that could potentially be used beyond namespaces for some other kind of resource in the future.

Using beyond namespaces doesn't preclude them from being in the namespace package

> There is a possibility we might need more than what we currently have for a region. Say a customer's private cloud region.

If this is expected to grow significantly, then no prob. In general, you want to avoid package sprawl, so seeing a single-message package is a bit offputting

> i am willing to move it to the namespace package if you feel strongly

I do not feel strongly here, what is currently here is ok too

PR #14, opened 2023-12-04
Account get and udpate apis

anekkanti:


temporal/api/cloud/account/v1/message.proto
```diff
@@ -0,0 +1,38 @@
...
+package temporal.api.cloud.account.v1;
+
+option go_package = "go.temporal.io/api/cloud/account/v1;account";
+
+message MetricsSpec {
+    // Enables the endpoint from which clients can scrape all their namespace metrics.
+    bool enabled = 1;
+    // The base64 encoded ca cert(s) in PEM format that clients connecting to the metrics endpoint can use for authentication.
+    // This must only be one value, but the CA can have a chain.
+    string accepted_client_ca = 2;
```
cretz, line 12:
```suggestion
    bytes accepted_client_ca = 2;
```

Why not just accept the PEM instead of base64 encoded (which is what proto JSON will be anyways FWIW)

temporal/api/cloud/account/v1/message.proto
```diff
@@ -0,0 +1,38 @@
...
+    string uri = 1;
+}
+
+message Account {
+    // The current version of the account specification.
+    // The next update operation will have to include this version.
+    string resource_version = 1;
+    // The account specification
+    AccountSpec spec = 2;
+    // The current state of the account.
```
cretz, line 32: If we aren't going to use proper enumerates, can we at least document all possible values?

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -62,60 +62,75 @@ service CloudService {
...
             get: "/api/v1/regions",
         };
     }
 
     // Get a region
     rpc GetRegion (GetRegionRequest) returns (GetRegionResponse) {
-	option (google.api.http) = {
-            get: "/api/v1/regions/{region}",
+        option (google.api.http) = {
+            get: "/api/v1/{region}",
```
```suggestion
            get: "/api/v1/regions/{region}",
```

Accidentally remove this?

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -229,3 +230,28 @@ message GetRegionResponse {
     // The temporal cloud region.
     temporal.api.cloud.region.v1.Region region = 1;
 }
+
+message GetAccountRequest {
```
cretz, line 234: Do we ever expect that a user can view/edit users from another account? Often in these cases you want to accept an identifier and just default it to the current user account. This allows you to prepare for a future general purpose API. You can still provide two HTTP endpoints, one for identifiable accounts resources and one for current.

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -62,60 +62,75 @@ service CloudService {
...
+        };
+    }
+
+    // Get account information
+    rpc GetAccount (GetAccountRequest) returns (GetAccountResponse) {
+            option (google.api.http) = {
```
cretz, line 124:
```suggestion
        option (google.api.http) = {
```

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -62,60 +62,75 @@ service CloudService {
...
+            get: "/api/v1/account",
+        };
+    }
+
+    // Update account information
+    rpc UpdateAccount (UpdateAccountRequest) returns (UpdateAccountResponse) {
+            option (google.api.http) = {
```
cretz, line 131:
```suggestion
        option (google.api.http) = {
```

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -229,3 +230,28 @@ message GetRegionResponse {
...
+    // The account.
+    temporal.api.cloud.account.v1.Account account = 1;
+}
+
+message UpdateAccountRequest {
+    // The updated account specification to apply.
+    temporal.api.cloud.account.v1.AccountSpec spec = 1;
+    // The version of the account for which this update is intended for.
+    // The latest version can be found in the GetAccount operation response.
+    string resource_version = 2;
```
cretz, line 247: Doesn't matter much, but was a little offputting to see the order of these two fields is different than it is in the account object itself

temporal/api/cloud/account/v1/message.proto
anekkanti: listed all the possible states

temporal/api/cloud/cloudservice/v1/service.proto
anekkanti: fixed

temporal/api/cloud/account/v1/message.proto
```diff
@@ -0,0 +1,40 @@
...
+}
+
+message Metrics {
+    // The prometheus metrics endpoint uri.
+    // This is only populated when the metrics is enabled in the metrics specification.
+    string uri = 1;
+}
+
+message Account {
+    // The account specification
```
cretz, line 27:
```suggestion
    // The account specification.
```

Here and elsewhere for consistency (though I know that we are probably inconsistent in api repo on trailing punctuation)

temporal/api/cloud/account/v1/message.proto
```diff
@@ -0,0 +1,40 @@
...
+    AccountSpec spec = 1;
+    // The current version of the account specification.
+    // The next update operation will have to include this version.
+    string resource_version = 2;
+    // The current state of the account.
+    // Possible values: activating, activationfailed, active, updating, updatefailed, deleting, deletefailed, deleted, suspending, suspendfailed, suspended
```
cretz, line 33:
```suggestion
    // Possible values: activating, activationfailed, active, updating, updatefailed, deleting, deletefailed, deleted, suspending, suspendfailed, suspended.
```

Many proto documentation utilities will not consider a newline to be end of sentence and so this will run on to the next.

temporal/api/cloud/account/v1/message.proto
cretz: This was my fault for not checking that well enough. PEM is already mostly base64, it's a bit weird to double encode. But I guess it's up to y'all.

temporal/api/cloud/account/v1/message.proto
anekkanti: thanks, fixed.

temporal/api/cloud/account/v1/message.proto
anekkanti: thanks fixed.

temporal/api/cloud/cloudservice/v1/request_response.proto
cretz: I think there is still value for consistency sake to accept an identifier here and expose an identifier on `Account`. The identifier here can default to the current user's.

anekkanti: I am thinking we can easily add the account-id later when we need to. Something like this: 
https://github.com/temporalio/api-cloud/compare/abhinav/accountManagement...abhinav/accountManagement-acctIDExt => adds a account-id to a few endpoints.

cretz: You definitely _can_ wait, I was just saying I think the consistency across resources has value now. At the very least even if not adding to the getter, maybe add an `id` to `Account`.

anekkanti: > maybe add an id to Account

This has had a lot of conversations on the cloud team. Over the last 2 years we have made a lot of effort to hide the account-id from the user. This was one of the product requirements, "a customer should not need to deal with their account ids...", there is some legacy reasons associated to it.

cretz: I think we should differentiate hiding from a standard user and hiding from a developer. If we want to hide from both, ok, but I disagree and I think an account identifier (even if you can't imagine how it could be used today) is a normal thing to have.

temporal/api/cloud/account/v1/message.proto
tminusplus: If we keep the double-encoding, we should add to the comment about this. Regardless of bytes or strings, I would have expected that `base64 encoded ca cert(s)` means PEM format.

temporal/api/cloud/account/v1/message.proto
```diff
@@ -0,0 +1,40 @@
...
+}
+
+message Account {
+    // The account specification.
+    AccountSpec spec = 1;
+    // The current version of the account specification.
+    // The next update operation will have to include this version.
+    string resource_version = 2;
+    // The current state of the account.
+    // Possible values: activating, activationfailed, active, updating, updatefailed, deleting, deletefailed, deleted, suspending, suspendfailed, suspended.
```
mastermanu, line 33: remember, let's remove suspendfailed, suspending from here

temporal/api/cloud/account/v1/message.proto
mastermanu: if we don't want to change the namespace spec (since I guess it's public preview), the other alternative is to just add another field which takes bytes to both account/namespace and then eventually deprecate the string field?

mastermanu (approved):

temporal/api/cloud/account/v1/message.proto
anekkanti: Okay, changed the type to []byte. Which expects the raw pem certs.

temporal/api/cloud/cloudservice/v1/request_response.proto
anekkanti: Added id to the account as suggested.

cretz: I think it would make sense for this operation to accept an account ID, but we can table it for the future. If set now, it can be defaulted today to requiring it either be unset or set as the current caller's account ID.

temporal/api/cloud/identity/v1/message.proto
```diff
@@ -60,6 +60,8 @@ message User {
     // The user specification
     UserSpec spec = 3;
     // The current state of the user
+    // Possible values: activating, activationfailed, active, updating, updatefailed, deleting, deletefailed, deleted, suspending, suspendfailed, suspended.
+    // For any failed state, reach out to Temporal Cloud support for remediation.
     string state = 4;
```
cretz, line 65: We should consider making this an enum, but up to y'all.

captainbeardo (approved):

temporal/api/cloud/identity/v1/message.proto
anekkanti: This is becoming an enum in a different PR.

temporal/api/cloud/cloudservice/v1/request_response.proto
anekkanti: I am inclining to adding it in the future. Might create a little unnecessary confusion with what to use for this optional field.

PR #16, opened 2024-03-12
Cloud API: Nexus Incoming/Outgoing Service Definitions

mastermanu:


temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -118,4 +118,41 @@ service CloudService {
             get: "/api/v1/regions/{region}",
         };
     }
+
+    // Gets all known nexus incoming services
+    rpc GetIncomingServices(GetIncomingServicesRequest) returns (GetIncomingServicesResponse) {
+	option (google.api.http) = {
+            get: "/api/v1/nexus/incomingservices",
```
mastermanu, line 125: Open Question: Given incoming services are already scoped to an account, is there a need to have this "incoming services" path here? This felt more forward-facing, but can understand the argument to just have these hang off the top-level nexus path.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,46 @@
+syntax = "proto3";
+
+package temporal.api.cloud.nexus.v1;
+
+option go_package = "go.temporal.io/api/cloud/nexus/v1;nexus";
+
+import "google/protobuf/timestamp.proto";
+
+message IncomingServiceSpec {
+    // The name of the incoming service. Must be unique within an account
```
mastermanu, line 10: Open Question:

For OSS, the "name" factors into the URL used when tying to access the service from a cluster.

For Cloud, the URL for addressing of services would be an internal implementation detail and generated by the server. In other words, we aren't using the incoming service name for anything here, so technically, we 1) do not need to enforce uniqueness and 2) this could be a friendly name/description instead. So should we still follow the same semantics here? We can do so just for consistency, but want to make sure there is "value" to the customer here.
3) Can we get away with making this immutable?

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,46 @@
...
+
+    // The namespace to route requests to
+    string namespace = 2;
+
+    // The task queue to route requests to
+    string task_queue = 3;
+}
+
+// A binding from a service name to namespace and task queue for dispatching incoming Nexus requests.
+message IncomingService {
```
mastermanu, line 22: no metadata map here. can be added later.

temporal/api/cloud/nexus/v1/message.proto
mastermanu: conversely, we can require the outgoing service reference this name directly.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,46 @@
...
+    // The id of the service. This is generated by the server and is immutable
+    string id = 1;
```
mastermanu, line 24: 1) so there is "id" and there is "name". Do we need both for cloud? Does outgoing service reference "id" or "name" in the API (note: UX can still "hide" id concept from customer if we want)

2) Should we add an "endpoints" section? Or is namespace_name/task_queue enough to derive the appropriate information?

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,46 @@
...
+
+    // The current version of the incoming service specification
+    // The next update operation must include this version
+    string resource_version = 2;
```
mastermanu, line 28: resource_version is control plane's flavor of "version/ETag", is a string and is consistent with other resources. This is another difference between this and the core OSS API

temporal/api/cloud/cloudservice/v1/service.proto
bergundy: I think it's better to be explicit and leave the `/nexus/` path open for other endpoints.

temporal/api/cloud/nexus/v1/message.proto
bergundy: No need for this in Cloud now and maybe never.

temporal/api/cloud/nexus/v1/message.proto
bergundy: We don't expose ID in the OSS API, I think we should stick with that in Cloud too for consistency.

All of the OSS endpoints use names.

temporal/api/cloud/nexus/v1/message.proto
bergundy: That's fine IMHO.

temporal/api/cloud/nexus/v1/message.proto
mastermanu: We can stick with just name, the only ask then is that the name is immutable.

The advantages of a server generated ID are:
- name can be mutable
- the ID we generate can eventually be referenced cross-account without any changes (future facing)
- Other artifacts in cloud (e.g. users) have a generated ID that is used as the primary key, so this is for consistency.
- We have severely regretted having the "user-name" tie into cloud resources in the past.

Am open to either option here, just want to discuss once beforehand

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+    string name = 1;
+
+    // The namespace to route requests to
+    string namespace = 2;
+
+    // The task queue to route requests to
+    string task_queue = 3;
+
+    // The authorization specification for this service
+    IncomingServiceAuthorizationSpec authorization_spec = 4;
```
mastermanu, line 21: TODO: Discuss if this is how we want to do it?

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,93 @@
...
+}
+
+message IncomingServiceAuthorizationSpec {
+    // The set of namespaces that are allowed to call into this service
+    repeated string namespaces = 1;
```
mastermanu, line 26: should we have a nested structure and have "Allowlist" as one of the options for future compatibility ?

temporal/api/cloud/nexus/v1/message.proto
mastermanu: note: if we do go the ID route, we can still try a) add GetServiceByName and b) enforce uniqueness

mastermanu: TL;DR, options:

1) name only, everything goes through name. name is immutable. outgoing services reference name
2) name + id. require name uniqueness in account still. outgoing services reference id
3) name + id. no requirements on uniqueness. outgoing services reference id

OSS currently does 1. There may be benefits for 2/3 that we should discuss before we choose.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,101 @@
...
+    string name = 1;
+
+    BindingSpec spec = 2;
+}
+
+message BindingSpec {
+    IntraAccountServiceSpec intra_account_service_spec = 1;
+}
+
+message IntraAccountServiceSpec {
```
mastermanu, line 72: in the future we could add GenericURI spec or CrossAccountServiceSpec of w/e else?

temporal/api/cloud/nexus/v1/message.proto
mastermanu: discussed offline, will go with approach 2.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,101 @@
...
+
+    // The task queue to route requests to
+    string task_queue = 3;
+
+    // The authorization specification for this service
+    IncomingServiceAuthorizationSpec authorization_spec = 4;
+}
+
+message IncomingServiceAuthorizationSpec {
+    AllowedNamespacePolicySpec allowed_namespace_policy_spec = 1;
```
mastermanu, line 25: [Abhinav] - Add word "Cloud" in there somewhere

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,101 @@
...
+
+    // The date and time when the service was last modified
+    // Will not be set if the service has never been modified.
+    google.protobuf.Timestamp last_modified_time = 7;
+}
+
+message OutgoingServiceSpec {
+    // The name of the outgoing service. Must be unique within the namespace
+    // The name must match `[a-zA-Z_][a-zA-Z0-9_]*`.
+    // This name is immutable
```
mastermanu, line 62: Open Question: do we also want id here?

temporal/api/cloud/nexus/v1/message.proto
mastermanu: IntraCloudServiceSpec ?

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,101 @@
...
+    string name = 1;
+
+    BindingSpec spec = 2;
+}
+
+message BindingSpec {
+    IntraAccountServiceSpec intra_account_service_spec = 1;
```
mastermanu, line 69: "one of" logic here?

Target instead of Binding or `BindingTarget` or `TargetDestination`

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -229,3 +230,187 @@ message GetRegionResponse {
...
+    // Cannot exceed 1000. Defaults to 100. 
+    int32 page_size = 1;
+
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+
+    // Filter incoming services by the namespace they route to - optional.
+    string namespace = 3;
+
+    // Filter incoming services by the task queue they route to - optional.
```
anekkanti, line 245: nit: Do we really need to support filtering with task_queue? Can't think of a usecase.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+    string id = 1;
+
+    // The current version of the incoming service specification
+    // The next update operation must include this version
+    string resource_version = 2;
+
+    // The incoming service specification
+    IncomingServiceSpec spec = 3;
+
+    // The current state of the service
```
anekkanti, line 51: List all the possible states of the service.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+    // Possible values are: ALLOWED_CLOUD_NAMESPACE
+    string policy_type = 1;
+
+    // A policy spec that allows one caller namespace to access the service
+    AllowedCloudNamespacePolicySpec allowed_cloud_namespace_policy_spec = 2;
+}
+
+message AllowedCloudNamespacePolicySpec {
+    // The namespace that is allowed to call into this service. Calling namespace must be in same account as incoming service
+    string namespace = 1;
```
anekkanti, line 36: Just to be clear, this is the complete namespace string (abhinavns1.a2dd6) and not just namespace name (abhinavns1)?

If yes, we won't need to change a thing when we start supporting cross account services right?

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+option go_package = "go.temporal.io/api/cloud/nexus/v1;nexus";
+
+import "google/protobuf/timestamp.proto";
+
+message IncomingServiceSpec {
+    // The name of the incoming service. Must be unique within an account
+    // The name must match `[a-zA-Z_][a-zA-Z0-9_]*`.
+    string name = 1;
+
+    // The namespace to route requests to
```
anekkanti, line 14: Can we add a comment for the spec fields about their mutability.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+    string namespace = 2;
+
+    // The task queue to route requests to
+    string task_queue = 3;
+
+    // The set of authorization policies for the service. Each request's caller
+    // must match with at least one of the specs to be accepted by the incoming service
+    repeated IncomingServiceAuthorizationSpec authorization_spec = 4;
```
anekkanti, line 22: Does the ordering of these specs mean anything? If yes, we need to make sure we can retrieve the same ordering that the user provides.

anekkanti, line 22:
```suggestion
    repeated IncomingServiceAuthorizationSpec authorization_specs = 4;
```

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+message IncomingServiceAuthorizationSpec {
+    // The policy set on this incoming service authorization spec
+    // Possible values are: ALLOWED_CLOUD_NAMESPACE
+    string policy_type = 1;
+
+    // A policy spec that allows one caller namespace to access the service
```
anekkanti, line 30:
```suggestion
    // A policy spec that allows one caller namespace to access the service.
    // Expected to be set when the policy_type is ALLOWED_CLOUD_NAMESPACE.
```

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+    string task_queue = 3;
+
+    // The set of authorization policies for the service. Each request's caller
+    // must match with at least one of the specs to be accepted by the incoming service
+    repeated IncomingServiceAuthorizationSpec authorization_spec = 4;
+}
+
+message IncomingServiceAuthorizationSpec {
+    // The policy set on this incoming service authorization spec
+    // Possible values are: ALLOWED_CLOUD_NAMESPACE
```
anekkanti, line 27: can we do `allowed_cloud_namespace` instead?

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+message TargetSpec {
+    // The type of target that is set in this spec
+    // Possible values are: CLOUD_INCOMING_SERVICE_SPEC
+    string target_type = 1;
+    
+    // A target spec for an incoming service within the same account as the outgoing service
+    CloudIncomingServiceSpec cloud_incoming_service_spec = 2;
+}
+
+message CloudIncomingServiceSpec {
```
anekkanti, line 84:
```suggestion
message TargetCloudIncomingServiceSpec {
```
Otherwise its very confusing with the `IncomingServiceSpec` that we already have.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+    string name = 1;
+
+    // The current version of the incoming service specification
+    // The next update operation must include this version
+    string resource_version = 2;
+
+    // The incoming service specification
+    OutgoingServiceSpec spec = 3;
+
+    // The current state of the service
```
anekkanti, line 102: List all the possible states.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,114 @@
...
+}
+
+message CloudIncomingServiceSpec {
+    // The ID of the Incoming Service that the outgoing service will connect to
+    string incoming_service_id = 1;
+}
+
+// A per-namespace binding from service name to destination service to invoke Nexus requests
+// that are initiated by workflows.
+message OutgoingService {
```
anekkanti, line 91: Will we ever need to support a use case where we need to list all the outgoing services in the account? if yes, we should add a namespace to the OutgoingService message.

anekkanti (commented):
Can we make sure to make all the comments full sentences. Meaning starts with a capitalized word and ends with a period.

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: My thought was that we might multiple services across different environments (test, stage, dev, prod) where the task queue is the same, but the namespace varies. @bergundy @prasek if this makes sense or not

temporal/api/cloud/nexus/v1/message.proto
mastermanu: will do. all of these are mutable

temporal/api/cloud/nexus/v1/message.proto
mastermanu: i think the ordering doesn't matter here (minus any form of short-circuit optimization), but cc @prasek @bergundy if you have thoughts

temporal/api/cloud/nexus/v1/message.proto
mastermanu: yes, this is the complete namespace string

temporal/api/cloud/nexus/v1/message.proto
mastermanu: good question. I think that might be too much (sort of like, listing all search attributes across all namespaces, or listing all export sinks across all namespaces). But agree it doesn't hurt to add the namespace to the message, let me do that

temporal/api/cloud/nexus/v1/message.proto
mastermanu: the other case where we do this "states" uses all caps for values. Do we want to be consistent with that?

temporal/api/cloud/nexus/v1/message.proto
mastermanu: we aren't doing this for user or namespace. this also risks getting out of date if someone doesn't update it as we add more states

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,122 @@
...
+    // The current version of the incoming service specification.
+    // The next update operation must include this version.
+    string resource_version = 2;
+
+    // The incoming service specification.
+    IncomingServiceSpec spec = 3;
+
+    // The current state of the service.
+    // Possible values: activating, activationfailed, active, updating, updatefailed, deleting, deletefailed, deleted
+    // For any failed state, reach out to Temporal Cloud support for remediation.
```
mastermanu, line 54: mention eventual consistency here?

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: spoke offline, we think this is useful to have this filter here

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -229,3 +230,187 @@ message GetRegionResponse {
...
+    // The list of incoming services in ascending ids order
+    repeated temporal.api.cloud.nexus.v1.IncomingService incoming_services = 1;
+
+    // The next page's token
+    string next_page_token = 2;
+}
+
+message GetIncomingServiceRequest {
+    // The id of the incoming service to get
+    string incoming_service_id = 1;
```
bergundy, line 262: nit: can we just call this `id`? it's already in the context of an `GetIncomingServiceRequest`.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -229,3 +230,187 @@ message GetRegionResponse {
...
+
+message GetIncomingServicesResponse {
+    // The list of incoming services in ascending ids order
+    repeated temporal.api.cloud.nexus.v1.IncomingService incoming_services = 1;
+
+    // The next page's token
+    string next_page_token = 2;
+}
+
+message GetIncomingServiceRequest {
```
bergundy, line 260:
```suggestion
message GetIncomingServiceByIdRequest {
```

Just in case we want to add a "by name" request?
The list approach is also fine, just a bit harder to use IMHO.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -229,3 +230,187 @@ message GetRegionResponse {
...
+
+    // Filter incoming services by the task queue they route to - optional.
+    string task_queue = 4;
+
+    // Filter incoming services by their name - optional. Specifying this will result in zero or one results.
+    string name = 5;
+}
+
+message GetIncomingServicesResponse {
+    // The list of incoming services in ascending ids order
```
bergundy, line 253: Looks like you stopped using periods for sentences from here on down.
I'm very much in favor of always using periods in comments.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -229,3 +230,187 @@ message GetRegionResponse {
     // The temporal cloud region.
     temporal.api.cloud.region.v1.Region region = 1;
 }
+
+message GetIncomingServicesRequest {
```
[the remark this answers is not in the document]
bergundy: Please mention if these are AND or OR filters in the comment for this message.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -229,3 +230,187 @@ message GetRegionResponse {
...
+    int32 page_size = 1;
+
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+
+    // The namespace to get all the outgoing services from
+    string namespace = 3;
+
+    // The name of the outgoing service to filter on - optional. Specifying this will result in zero or one results.
+    string name = 4;
```
bergundy, line 335: Not really useful as a filter, this is basically a Get request.

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -118,4 +118,78 @@ service CloudService {
             get: "/api/v1/regions/{region}",
         };
     }
+
+    // Gets all known incoming services
+    rpc GetIncomingServices(GetIncomingServicesRequest) returns (GetIncomingServicesResponse) {
+	option (google.api.http) = {
+            get: "/api/v1/nexus/incomingservices",
```
bergundy, line 125: Can we add a `/by-id` or `/id` suffix to this URL just in case we want to allow looking up by name eventually?

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,122 @@
...
+
+    // The set of authorization policies for the service. Each request's caller
+    // must match with at least one of the specs to be accepted by the incoming service.
+    repeated IncomingServiceAuthorizationSpec authorization_specs = 4;
+}
+
+message IncomingServiceAuthorizationSpec {
+    // The policy set on this incoming service authorization spec.
+    // Possible values are: allowed_cloud_namespace
+    string policy_type = 1;
```
bergundy, line 29: Why not make this an enum, I'm curious?

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,122 @@
...
+    // The date and time when the service was last modified.
+    // Will not be set if the service has never been modified.
+    google.protobuf.Timestamp last_modified_time = 7;
+}
+
+message OutgoingServiceSpec {
+    // The name of the outgoing service. Must be unique within the namespace.
+    // The name must match `[a-zA-Z_][a-zA-Z0-9_]*`.
+    // This name is immutable.
+    string name = 1;
```
bergundy, line 72: This is the identifier, doesn't belong in the spec AFAICT.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// A per-namespace binding from service name to destination service to invoke Nexus requests
+// that are initiated by workflows.
+message OutgoingService {
+    // The name of the outgoing service.
+    string name = 1;
+
+    // The namespace this outgoing service is attached to.
+    string namespace = 2;
```
bergundy, line 99: Should this be `= 1`? It's essentially the first part of the identifier.

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: just an accident

temporal/api/cloud/nexus/v1/message.proto
mastermanu: @anekkanti can comment here. We appear to have stayed away from enums elsewhere (and are using strings instead), so this is just being consistent with the existing cloud APIs

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: right, this is what we did for our Users api (users have a static id, but you call GetUsersRequset and can supply an email address), so this is just an attempt to stay consistent.

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: we can. I was trying to be consistent with this one: https://github.com/temporalio/api-cloud/blob/main/temporal/api/cloud/cloudservice/v1/request_response.proto#L33

temporal/api/cloud/cloudservice/v1/request_response.proto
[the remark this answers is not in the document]
mastermanu: yeah the list vs get is for consistency with existing cloud apis. could have been list.

these are AND filters. will update comments

temporal/api/cloud/cloudservice/v1/service.proto
mastermanu: since I am following what we do for users, I've kept the same format (which doesn't have by-id or id either..). @anekkanti , thoughts?

temporal/api/cloud/nexus/v1/message.proto
mastermanu: good point. reordered

temporal/api/cloud/nexus/v1/message.proto
mastermanu: that seems to be what we have done for namespace and for outgoing service. it is a property a user can set (albeit only on create), so we happened to put it on spec. can try and move it off, but it does create some inconsistency with the other cloud apis

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: @anekkanti - thoughts? we basically followed the users model, which relies on this

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: agreeing with sergey here. am inclined towards leaving this as is for now (e.g. GetIncomingServiceRequest) just to be consistent with existing cloud user apis

the weirdness is that when we translate this to HTTP path, we would end up creating a different path at that time probably, but not the end of the world if we have to introduce that

temporal/api/cloud/cloudservice/v1/service.proto
mastermanu: yes, am inclined to leave as-is as well. can always add suffix later if needed

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,122 @@
...
+    string state = 4;
+
+    // The id of the async operation that is creating/updating/deleting the service, if any.
+    string async_operation_id = 5;
+
+    // The date and time when the service was created.
+    google.protobuf.Timestamp created_time = 6;
+
+    // The date and time when the service was last modified.
+    // Will not be set if the service has never been modified.
```
bergundy, line 64: Curious why not default this to `created_time`? Is this for consistency with other APIs?

mastermanu: yeah, this is just for consistency. could have definitely been defaulted to created_time here

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,122 @@
...
+
+    // The namespace to route requests to. This field is mutable.
+    string namespace = 2;
+
+    // The task queue to route requests to. This field is mutable.
+    string task_queue = 3;
+
+    // The set of authorization policies for the service. Each request's caller
+    // must match with at least one of the specs to be accepted by the incoming service.
+    repeated IncomingServiceAuthorizationSpec authorization_specs = 4;
```
bergundy, line 23: I think maybe instead of auth spec, we might want to have a general message to also add limits (rate, concurrency, etc..) and other properties for every caller (only namespaces for now).

cretz (approved):
I have no real comments/opinions on this one. From a high-level LGTM, but would defer to those with more familiarity with the behavior.

mastermanu:
abandoning as there is replacement PR

PR #19, opened 2024-04-02
Add Cloud public API for multi region namespace

yux0:
## What was changed
Add Cloud public API for multi region namespace

## Why?
Add Cloud public API for multi region namespace

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -227,23 +227,6 @@ message FailoverNamespaceResponse {
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
-message GetNamespaceFailoverHistoryRequest {
-    // The requested size of the page to retrieve - optional.
-    // Cannot exceed 1000. Defaults to 100.
-    int32 page_size = 1;
-    // The page token if this is continuing from another response - optional.
-    string page_token = 2;
-    // The namespace to get failover history.
-    string namespace = 3;
-}
-
-message GetNamespaceFailoverHistoryResponse {
-    // The page token if this is continuing from another response
-    string page_token = 1;
-    // Namespace failover history
-    repeated temporal.api.cloud.namespace.v1.FailoverRecord failover_history = 2;
-}
-
 message AddNamespaceRegionRequest {
     // The namespace to failover.
     string namespace = 1;
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -212,6 +212,53 @@ message DeleteNamespaceResponse {
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
+message FailoverNamespaceRequest {
+    // The namespace to failover.
+    string namespace = 1;
+    // The target region to failover.
+    temporal.api.cloud.region.v1.Region target_region = 2;
+    // The id to use for this async operation.
+    // Optional, if not provided a random id will be generated.
+    string async_operation_id = 3;
+}
+
+message FailoverNamespaceResponse {
+    // The async operation.
+    temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
+}
+
+message GetNamespaceFailoverHistoryRequest {
+    // The requested size of the page to retrieve - optional.
+    // Cannot exceed 1000. Defaults to 100.
+    int32 page_size = 1;
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+    // The namespace to get failover history.
+    string namespace = 3;
+}
+
+message GetNamespaceFailoverHistoryResponse {
+    // The page token if this is continuing from another response
+    string page_token = 1;
+    // Namespace failover history
+    repeated temporal.api.cloud.namespace.v1.FailoverRecord failover_history = 2;
+}
+
+message AddNamespaceRegionRequest {
+    // The namespace to failover.
+    string namespace = 1;
+    // The new region to be added.
+    temporal.api.cloud.region.v1.Region new_region = 2;
+    // The id to use for this async operation.
+    // Optional, if not provided a random id will be generated.
+    string async_operation_id = 3;
+}
+
+message AddNamespaceRegionResponse {
+    // The async operation.
+    temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
+}
+
 message GetRegionsRequest {
 }
 
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -215,8 +215,8 @@ message DeleteNamespaceResponse {
 message FailoverNamespaceRequest {
     // The namespace to failover.
     string namespace = 1;
-    // The target region to failover.
-    temporal.api.cloud.region.v1.Region target_region = 2;
+    // The target region id to failover.
+    string target_region_id = 2;
     // The id to use for this async operation.
     // Optional, if not provided a random id will be generated.
     string async_operation_id = 3;
@@ -247,8 +247,8 @@ message GetNamespaceFailoverHistoryResponse {
 message AddNamespaceRegionRequest {
     // The namespace to failover.
     string namespace = 1;
-    // The new region to be added.
-    temporal.api.cloud.region.v1.Region new_region = 2;
+    // The new region id to be added.
+    string new_region_id = 2;
     // The id to use for this async operation.
     // Optional, if not provided a random id will be generated.
     string async_operation_id = 3;
--- a/temporal/api/cloud/cloudservice/v1/service.proto
+++ b/temporal/api/cloud/cloudservice/v1/service.proto
@@ -113,13 +113,6 @@ service CloudService {
         };
     }
 
-    // Get failover history on a namespace
-    rpc GetNamespaceFailoverHistory (GetNamespaceFailoverHistoryRequest) returns (GetNamespaceFailoverHistoryResponse) {
-        option (google.api.http) = {
-            get: "/api/v1/namespaces/{namespace}/failover-history",
-        };
-    }
-
     // Add a new region to a namespace
     rpc AddNamespaceRegion (AddNamespaceRegionRequest) returns (AddNamespaceRegionResponse) {
         option (google.api.http) = {
--- a/temporal/api/cloud/cloudservice/v1/service.proto
+++ b/temporal/api/cloud/cloudservice/v1/service.proto
@@ -105,6 +105,29 @@ service CloudService {
         };
     }
 
+    // Failover a multi-region namespace
+    rpc FailoverNamespace (FailoverNamespaceRequest) returns (FailoverNamespaceResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/failover",
+            body: "*"
+        };
+    }
+
+    // Get failover history on a namespace
+    rpc GetNamespaceFailoverHistory (GetNamespaceFailoverHistoryRequest) returns (GetNamespaceFailoverHistoryResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}/failover-history",
+        };
+    }
+
+    // Add a new region to a namespace
+    rpc AddNamespaceRegion (AddNamespaceRegionRequest) returns (AddNamespaceRegionResponse) {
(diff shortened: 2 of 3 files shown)


temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -105,6 +105,29 @@ service CloudService {
...
+
+    // Get failover history on a namespace
+    rpc GetNamespaceFailoverHistory (GetNamespaceFailoverHistoryRequest) returns (GetNamespaceFailoverHistoryResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/namespaces/{namespace}/failover-history",
+        };
+    }
+
+    // Add a new region to a namespace
+    rpc AddNamespaceRegion (AddNamespaceRegionRequest) returns (AddNamespaceRegionResponse) {
```
yux0, line 124: Another option: Use UpdateNamespace to support this case.

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -105,6 +105,29 @@ service CloudService {
         };
     }
 
+    // Failover a multi-region namespace
+    rpc FailoverNamespace (FailoverNamespaceRequest) returns (FailoverNamespaceResponse) {
```
cretz, line 109: Users can trigger a failover at will? Or if this is internal-only API, can it be removed from the user-facing API? Same for `AddNamespaceRegion`.

yux0: Yes. They are user facing APIs. Customers can add region or failover from UI as well.

cretz: This is a bit surprising to me that we let users add regions and failover whenever they want but I admit to not knowing enough details about the project.

cretz (approved):

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -124,3 +125,25 @@ message Namespace {
     // Will not be set if the namespace has never been modified.
     google.protobuf.Timestamp last_modified_time = 11;
 }
+
+enum FailoverType {
```
anekkanti, line 129: We don't do enums on the cloud apis. 
We choose to use strings instead.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -124,3 +125,25 @@ message Namespace {
...
+
+enum FailoverStatus {
+    FAILOVER_STATUS_UNSPECIFIED = 0;
+    FAILOVER_STATUS_SUCCEEDED = 1;
+    FAILOVER_STATUS_FAILED = 2;
+}
+
+message FailoverRecord {
+    FailoverType failover_type = 1;
+    temporal.api.cloud.region.v1.Region source_region = 2;
```
anekkanti, line 143: Lets just use a string instead like the way we do on the namespace message. 
https://github.com/temporalio/api-cloud/blob/main/temporal/api/cloud/namespace/v1/message.proto#L48C4-L54C33

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -124,3 +125,25 @@ message Namespace {
...
+}
+
+enum FailoverStatus {
+    FAILOVER_STATUS_UNSPECIFIED = 0;
+    FAILOVER_STATUS_SUCCEEDED = 1;
+    FAILOVER_STATUS_FAILED = 2;
+}
+
+message FailoverRecord {
+    FailoverType failover_type = 1;
```
anekkanti, line 142: Can we add more comments to the these fields and what they mean.

temporal/api/cloud/namespace/v1/message.proto
anekkanti, line 143: Do we need a source_region? Isn't it implied by the active region of the namespace?
https://github.com/temporalio/api-cloud/blob/main/temporal/api/cloud/namespace/v1/message.proto#L116

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -105,6 +105,29 @@ service CloudService {
...
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/failover",
+            body: "*"
+        };
+    }
+
+    // Get failover history on a namespace
+    rpc GetNamespaceFailoverHistory (GetNamespaceFailoverHistoryRequest) returns (GetNamespaceFailoverHistoryResponse) {
```
anekkanti, line 117: Why do we need this? Doesn't the audit logs capture this?

temporal/api/cloud/cloudservice/v1/service.proto
yux0: @atihkin Is this ok to you?

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -212,6 +212,53 @@ message DeleteNamespaceResponse {
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
+message FailoverNamespaceRequest {
+    // The namespace to failover.
+    string namespace = 1;
+    // The target region to failover.
+    temporal.api.cloud.region.v1.Region target_region = 2;
```
anekkanti, line 219: I would just use a region_id like we do on the namespace spec.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -46,11 +47,10 @@ message NamespaceSpec {
     // The name is immutable. Once set, it cannot be changed.
     string name = 1;
     // The ids of the regions where the namespace should be available.
-    // Specifying more than one region makes the namespace "global", which is currently a preview only feature with restricted access.
+    // Specifying more than one region is a multi-region namespace.
     // Please reach out to Temporal support for more information on global namespaces.
     // When provisioned the global namespace will be active on the first region in the list and passive on the rest.
-    // Number of supported regions is 2. 
-    // The regions is immutable. Once set, it cannot be changed.
+    // Number of supported regions is 2.
```
yux0, line 53: @atihkin to review the doc

temporal/api/cloud/cloudservice/v1/service.proto
atihkin: - We have the ability to limit region pairs however we won't stop users from choosing regions that fit their infra needs (where their workers are, other dependent services)
- Users need the ability to trigger a failover for testing (to make sure they have set everything up correctly) and also for a break glass type scenario where they have an outage on their end that Temporal for some reason has not detected

temporal/api/cloud/cloudservice/v1/service.proto
atihkin: can we keep this capabaility in AddNamespaceRegion instead?

temporal/api/cloud/cloudservice/v1/request_response.proto
yux0: Should we migrate the namespace spec to Region? rather than move this to a string region id?

temporal/api/cloud/cloudservice/v1/service.proto
yux0: I think this is more from a debugging/dc stand point to get the failover history. But I don't have a strong opinion. @atihkin do you think exposing this is necessary or we can defer this.

temporal/api/cloud/namespace/v1/message.proto
yux0: same concern. If there is a plan to multi cloud. I think we should move to this Region object rather than use string region id.

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -124,3 +125,25 @@ message Namespace {
...
+}
+
+message FailoverRecord {
+    FailoverType failover_type = 1;
+    temporal.api.cloud.region.v1.Region source_region = 2;
+    temporal.api.cloud.region.v1.Region target_region = 3;
+    google.protobuf.Timestamp start_time_utc = 4;
+    google.protobuf.Timestamp end_time_utc = 5;
+    FailoverStatus status = 6;
+    string operator = 7;
```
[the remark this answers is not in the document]
yux0: yes. we have a place holder for internal operator.

temporal/api/cloud/cloudservice/v1/service.proto
yux0: Synced with @atihkin. We agreed to have this new API for two reasons:
1. Better user experience.
2. The add region is special and a longer async operation which is different than update namespace.

temporal/api/cloud/namespace/v1/message.proto
yux0: removed.

temporal/api/cloud/cloudservice/v1/service.proto
atihkin: We're showing this info in the UI so I think it should show up here as well.

meiliang86 (approved):

temporal/api/cloud/cloudservice/v1/service.proto
anekkanti: For automation use cases, the primary consumers of these apis, have a patch only endpoint adds complexity on the client. It will need to selectively call a AddNamespaceRegion() for updating the regions, why calling a UpdatingNamespace to update all else. 

We can have a additional patch endpoint. But i strongly believe UpdateNamespace should also support adding a region.

anekkanti: Also do we plan to add support for removing a region? anytime in the future.

temporal/api/cloud/cloudservice/v1/service.proto
anekkanti: These APIs are meant for automation use cases. Having parity with the UI is not really necessary. 
Though i am not really against having this specific API, but I would side on adding only if there is really a automation use case around it.

temporal/api/cloud/namespace/v1/message.proto
anekkanti: Multi cloud is addressed by the region id. We prefix the cloud type to the region. 

```
$ omni ocld cloud-apis regions list
{
	"regions": [
		{
			"id": "aws-us-east-1",
			"cloudProvider": "aws",
			"cloudProviderRegion": "us-east-1",
			"location": "US East (N. Virginia)"
		},
		{
			"id": "aws-us-west-2",
			"cloudProvider": "aws",
			"cloudProviderRegion": "us-west-2",
			"location": "US West (Oregon)"
		}
	]
}
```

temporal/api/cloud/cloudservice/v1/service.proto
yux0: Yes. in long term we will support.

temporal/api/cloud/cloudservice/v1/service.proto
yux0: @atihkin If the cloud api is for automation use cases, I don't think we want to support them when launching?

temporal/api/cloud/cloudservice/v1/service.proto
yux0: My concern is the UpdateNamespace with new region and other fields together would be hard to deal with. Any good option we can limit to only update one field at a time?

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -212,6 +212,53 @@ message DeleteNamespaceResponse {
...
+    // Optional, if not provided a random id will be generated.
+    string async_operation_id = 3;
+}
+
+message FailoverNamespaceResponse {
+    // The async operation.
+    temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
+}
+
+message GetNamespaceFailoverHistoryRequest {
```
yux0, line 230: Remove for now. Follow up with @jumbogo about expose those fields from audit logging.

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -112,6 +112,22 @@ service CloudService {
         };
     }
 
+    // Failover a multi-region namespace
+    rpc FailoverNamespace (FailoverNamespaceRequest) returns (FailoverNamespaceResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/failover",
```
```suggestion
            post: "/api/v1/cloud/namespaces/{namespace}/failover",
```

Sorry to come in late on this, but as part of #20 we changed this root to be `/api/v1/cloud`. Can you apply this root to the two calls here? Thanks!

PR #22, opened 2024-04-16
Introduce user group management APIs

codemonkeycxy:
## What was changed

Introducing a new identity concept "UserGroup" with CRUD APIs that manage user group -> account/namespace access mapping

## Why?

This is to enable the management of user group -> account/namespace access mapping by Temporal Cloud customers
- Customers manage the user -> group mapping on their end
- Customer sets up group -> account/namespace access mapping by calling these APIs
- When users log in, Temporal Cloud uses the groups in user's token to perform authz

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -244,7 +244,7 @@ message GetUserGroupsRequest {
     // Only return groups that grant access to the specified namespace - optional.
     string namespace = 3;
     // Filter groups by their type - optional.
-    // Allowed values ["google-workspace"].
+    // Allowed values ["google_workspace"].
     string group_type = 4;
     // Filter groups by their name - optional.
     string group_name = 5;
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -241,14 +241,17 @@ message GetUserGroupsRequest {
     int32 page_size = 1;
     // The page token if this is continuing from another response - optional.
     string page_token = 2;
+    // Filter groups by their connection type - optional.
+    // Allowed values ["google-workspace"].
+    string connection_type = 3;
     // Only return groups that grant access to the specified namespace - optional.
-    string namespace = 3;
+    string namespace = 4;
     // Filter groups by their name - optional.
-    string name = 4;
+    string name = 5;
 }
 
 message GetUserGroupsResponse {
-    // The list of groups in ascending name order.
+    // The list of groups in ascending ids order.
     repeated temporal.api.cloud.identity.v1.UserGroup groups = 1;
     // The next page's token.
     string next_page_token = 2;
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -243,11 +243,11 @@ message GetUserGroupsRequest {
     string page_token = 2;
     // Only return groups that grant access to the specified namespace - optional.
     string namespace = 3;
-    // Filter groups by their host type - optional.
+    // Filter groups by their type - optional.
     // Allowed values ["google-workspace"].
-    string host_type = 4;
+    string group_type = 4;
     // Filter groups by their name - optional.
-    string name = 5;
+    string group_name = 5;
 }
 
 message GetUserGroupsResponse {
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -235,7 +235,7 @@ message GetRegionResponse {
     temporal.api.cloud.region.v1.Region region = 1;
 }
 
-message GetGroupsRequest {
+message GetUserGroupsRequest {
     // The requested size of the page to retrieve - optional.
     // Cannot exceed 1000. Defaults to 100.
     int32 page_size = 1;
@@ -247,45 +247,45 @@ message GetGroupsRequest {
     string name = 4;
 }
 
-message GetGroupsResponse {
+message GetUserGroupsResponse {
     // The list of groups in ascending name order.
-    repeated temporal.api.cloud.identity.v1.Group groups = 1;
+    repeated temporal.api.cloud.identity.v1.UserGroup groups = 1;
     // The next page's token.
     string next_page_token = 2;
 }
 
-message GetGroupRequest {
+message GetUserGroupRequest {
     // The id of the group to get.
     string group_id = 1;
 }
 
-message GetGroupResponse {
+message GetUserGroupResponse {
     // The group.
-    temporal.api.cloud.identity.v1.Group group = 1;
+    temporal.api.cloud.identity.v1.UserGroup group = 1;
 }
 
-message CreateGroupRequest {
+message CreateUserGroupRequest {
     // The id of the group.
     string group_id = 1;
     // The spec for the group to create.
-    temporal.api.cloud.identity.v1.GroupSpec spec = 2;
+    temporal.api.cloud.identity.v1.UserGroupSpec spec = 2;
     // The id to use for this async operation.
     // Optional, if not provided a random id will be generated.
     string async_operation_id = 3;
 }
 
-message CreateGroupResponse {
+message CreateUserGroupResponse {
     // The id of the group that was created.
     string group_id = 1;
     // The async operation.
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 2;
 }
 
-message UpdateGroupRequest {
+message UpdateUserGroupRequest {
     // The id of the group to update.
     string group_id = 1;
     // The new group specification.
-    temporal.api.cloud.identity.v1.GroupSpec spec = 2;
+    temporal.api.cloud.identity.v1.UserGroupSpec spec = 2;
     // The version of the group for which this update is intended for.
     // The latest version can be found in the GetGroup operation response.
     string resource_version = 3;
@@ -294,12 +294,12 @@ message UpdateGroupRequest {
     string async_operation_id = 4;
 }
 
-message UpdateGroupResponse {
+message UpdateUserGroupResponse {
     // The async operation.
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
-message DeleteGroupRequest {
+message DeleteUserGroupRequest {
     // The id of the group to delete.
     string group_id = 1;
     // The version of the group for which this delete is intended for.
@@ -310,12 +310,12 @@ message DeleteGroupRequest {
     string async_operation_id = 3;
 }
 
-message DeleteGroupResponse {
+message DeleteUserGroupResponse {
     // The async operation.
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
-message SetGroupNamespaceAccessRequest {
+message SetUserGroupNamespaceAccessRequest {
     // The namespace to set permissions for.
     string namespace = 1;
     // The id of the group to set permissions for.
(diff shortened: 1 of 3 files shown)


temporal/api/cloud/identity/v1/message.proto
```diff
@@ -69,3 +69,27 @@ message User {
...
+message GroupSpec {
+    // The human-readable name of the group
+    string name = 1;
+    // The type of the group. e.g. "googleworkspace"
+    string type = 2;
+    // The access assigned to the group
+    temporal.api.cloud.identity.v1.Access access = 3;
+}
+
+message Group {
```
cretz, line 82:
```suggestion
message UserGroup {
```

Can be something else, but "group" I think is far too generic

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -125,4 +125,49 @@ service CloudService {
             get: "/api/v1/cloud/regions/{region}",
         };
     }
+
+    // Get all groups
+    rpc GetGroups (GetGroupsRequest) returns (GetGroupsResponse) {
```
cretz, line 130:
```suggestion
    rpc GetUserGroups (GetUserGroupsRequest) returns (GetUserGroupsResponse) {
```

Same comment here as on object, just "group" is far too generic

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -125,4 +125,49 @@ service CloudService {
...
+        option (google.api.http) = {
+            get: "/api/v1/groups",
```
cretz, line 132:
```suggestion
            get: "/api/v1/cloud/groups",
```

We now prefix control plane with common HTTP prefix so it can live side-by-side with dataplane on the same single endpoint (also change this to `user-groups` if we change the RPC name)

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +234,102 @@ message GetRegionResponse {
...
+
+message GetGroupsRequest {
+    // The requested size of the page to retrieve - optional.
+    // Cannot exceed 1000. Defaults to 100.
+    int32 page_size = 1;
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+    // Only return groups that grant access to the specified namespace - optional.
+    string namespace = 3;
+    // Filter groups by their name - optional
```
cretz, line 246: You are inconsistent with trailing punctuation

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -125,4 +125,49 @@ service CloudService {
...
+        option (google.api.http) = {
+            post: "/api/v1/groups",
+            body: "*"
+        };
+    }
+
+    // Update a group
+    rpc UpdateGroup (UpdateGroupRequest) returns (UpdateGroupResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/groups/{group_id}",
```
cretz, line 154: It is a bit confusing that I can't POST with my ID to create the thing even though it is a user-supplied ID. Consider combining create and update into one operation using resource version to differentiate.

temporal/api/cloud/identity/v1/message.proto
```diff
@@ -69,3 +69,27 @@ message User {
     // Will not be set if the user has never been modified.
     google.protobuf.Timestamp last_modified_time = 8;
 }
+
+message GroupSpec {
+    // The human-readable name of the group
```
anekkanti, line 74: Isn't this more than just a human-readable name? Its the name of the external group, in this case a the name of the google workspace group.
Also i would add a comment stating mutable. I am assuming its not.

temporal/api/cloud/identity/v1/message.proto
anekkanti, line 82: I don't see a `status` field. The one that changes to `updating` when the group is being updated?

temporal/api/cloud/cloudservice/v1/service.proto
anekkanti, line 132: The rest paths need an update.
Now we do `/api/v1/cloud/...`

```suggestion
            get: "/api/v1/cloud/groups",
```

temporal/api/cloud/identity/v1/message.proto
```diff
@@ -69,3 +69,27 @@ message User {
...
+    string name = 1;
+    // The type of the group. e.g. "googleworkspace"
+    string type = 2;
```
anekkanti, line 77: Please specify the mutability of field in the comments.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +234,102 @@ message GetRegionResponse {
...
+    string group_id = 1;
+}
+
+message GetGroupResponse {
+    // The group.
+    temporal.api.cloud.identity.v1.Group group = 1;
+}
+
+message CreateGroupRequest {
+    // The id of the group
```
anekkanti, line 268:
```suggestion
    // The id of the group.
```

temporal/api/cloud/identity/v1/message.proto
codemonkeycxy: added state and async_operation_id

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -125,4 +125,49 @@ service CloudService {
             get: "/api/v1/cloud/regions/{region}",
         };
     }
+
+    // Get all groups
+    rpc GetGroups (GetUserGroupsRequest) returns (GetUserGroupsResponse) {
```
cretz, line 130:
```suggestion
    rpc GetUserGroups (GetUserGroupsRequest) returns (GetUserGroupsResponse) {
```

Need to change the RPC names here too

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -125,4 +125,49 @@ service CloudService {
...
+    rpc GetUserGroup (GetUserGroupRequest) returns (GetUserGroupResponse) {
+        option (google.api.http) = {
+            get: "/api/v1/cloud/user-groups/{group_id}",
+        };
+    }
+
+    // Create new a group
+    rpc CreateUserGroup (CreateUserGroupRequest) returns (CreateUserGroupResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/cloud/user-groups",
```
codemonkeycxy, line 146: Comments from @cretz
>It is a bit confusing that I can't POST with my ID to create the thing even though it is a user-supplied ID. Consider combining create and update into one operation using resource version to differentiate.

so we make a distinction between `group_id` and `group_name`. the group id is an uuid generated by us, the group_name is a string supplied by the customer.

when creating a user group, the customer can pass in the group name as part of the UserGroupSpec

cretz: Works for me and seems to be inline w/ what we do for users

cretz: How does the create/update of user work? Can the user specify the ID? Should the ID be on the spec?

cretz: Confirmed off-PR that this works just like user creation/update

cretz (approved):

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +234,100 @@ message GetRegionResponse {
...
 }
+
+message GetUserGroupsRequest {
+    // The requested size of the page to retrieve - optional.
+    // Cannot exceed 1000. Defaults to 100.
+    int32 page_size = 1;
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+    // Only return groups that grant access to the specified namespace - optional.
+    string namespace = 3;
```
codemonkeycxy, line 245: TODO: add a type filter

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +234,100 @@ message GetRegionResponse {
...
+    // Filter groups by their name - optional.
+    string name = 4;
+}
+
+message GetUserGroupsResponse {
+    // The list of groups in ascending name order.
```
codemonkeycxy, line 251: consider changing to ascending id order

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +234,103 @@ message GetRegionResponse {
...
+    // The requested size of the page to retrieve - optional.
+    // Cannot exceed 1000. Defaults to 100.
+    int32 page_size = 1;
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+    // Filter groups by the namespace they have access to - optional.
+    string namespace = 3;
+    // Filter groups by their type - optional.
+    // Allowed values ["google_workspace"].
+    string group_type = 4;
```
mattkim, line 248: Hmmmm why do we need this? Isn't this sort of invisible to them?

temporal/api/cloud/identity/v1/message.proto
```diff
@@ -66,6 +66,37 @@ message User {
...
 }
+
+message UserGroupSpec {
+    // The name of the group as defined in the customer's IdP (e.g. Google group name in Google Workspace)
+    // The name is immutable. Once set, it cannot be changed
+    string name = 1;
+    // The type of the group.
+    // This field is immutable. Once set, it cannot be changed
+    // Allowed values ["google_workspace"].
+    string type = 2;
```
mattkim, line 80: same comment

temporal/api/cloud/identity/v1/message.proto
```diff
@@ -66,6 +66,37 @@ message User {
...
     // The date and time when the user was last modified
-    // Will not be set if the user has never been modified.
+    // Will not be set if the user has never been modified
     google.protobuf.Timestamp last_modified_time = 8;
 }
+
+message UserGroupSpec {
+    // The name of the group as defined in the customer's IdP (e.g. Google group name in Google Workspace)
+    // The name is immutable. Once set, it cannot be changed
+    string name = 1;
```
mattkim, line 76: i wonder if we need a name and an "internal id"

codemonkeycxy: yes there's an internal id, it's not on the spec object though because it's set by our system (just like user). you can find that id on the `UserGroup` struct

temporal/api/cloud/cloudservice/v1/request_response.proto
codemonkeycxy: we want this for extensibility. say a company was on google_workspace but wants to switch to some other IdP. we might need to support 2 sets of groups with the same name. group_type is there to differentiate them

temporal/api/cloud/identity/v1/message.proto
anekkanti, line 80: I wonder if just the type is enough... should we add a connection_id too ?
But not sure how the user can know the connection id

anekkanti: never mind we can, add support for connection id later.

anekkanti (approved):

temporal/api/cloud/cloudservice/v1/request_response.proto
mattkim: I was thinking do we need an abstract primary key for groups and then also maintain that group name + group type are unique?

My preference is to not name this "group type" since it's so generic. Instead something more descriptive like "idp_type" would. Though the other question I have is--is this supposed to eventually be used for temporal only groups?

mattkim (approved):

temporal/api/cloud/cloudservice/v1/request_response.proto
codemonkeycxy: discussed offline and will remove group type for now. we can always add back in later

PR #23, opened 2024-05-01
Adds Nexus Endpoint

mastermanu:


temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -125,4 +125,41 @@ service CloudService {
...
+    
+    // Get a nexus endpoint
+    rpc GetNexusEndpoint(GetNexusEndpointRequest) returns (GetNexusEndpointResponse) {
+	option (google.api.http) = {
+            get: "/api/v1/nexus/endpoints/{endpoint_id}",
+        };
+    }
+
+    // Create a nexus endpoint
+    rpc CreateIncomingService(CreateIncomingServiceRequest) returns (CreateIncomingServiceResponse) {
```
mastermanu, line 144: doh

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,85 @@
...
+    // Indicates where the endpoint should forward received nexus requests to. 
+    EndpointTargetSpec target_spec = 2;
+
+    // The set of authorization policies for the endpoint. Each request's caller
+    // must match with at least one of the specs to be accepted by the endpoint.
+    // This field is mutable.
+    repeated EndpointAuthorizationSpec authorization_specs = 3;
+
+    // Optional description. Can be in markdown for rendering in the UI
+    string description = 4;
```
mastermanu, line 24: delta from OSS: that is a temporal api payload.
we have a string here and user would need to take care of encryption/sensitive data out of band (or link to a website which does auth)

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +234,89 @@ message GetRegionResponse {
...
+message GetNexusEndpointsRequest {
+    // The requested size of the page to retrieve - optional.
+    // Cannot exceed 1000. Defaults to 100. 
+    int32 page_size = 1;
+
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+
+    // Filter endpoints based on their target specification - optional, treated as an AND if specified. Empty fields within the spec are treated as '*'
+    temporal.api.cloud.nexus.v1.EndpointTargetSpec = 3;
```
mastermanu, line 247: delta from OSS: developers probably want ways to filter out endpoints based on either namespace, task queue or both. OSS doesn't have this right now. 

Other option is that we cut this for now and filter on client-side until a later point in time

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,85 @@
...
+}
+
+message EndpointTargetSpec {
+    // The type of target that this endpoint routes Nexus requests to.
+    // Possible values are: task_queue
+    string target_type = 1;
```
mastermanu, line 30: note: this is essentially a poor person's version of `oneof`

temporal/api/cloud/cloudservice/v1/request_response.proto
bergundy, line 247:
```suggestion
    temporal.api.cloud.nexus.v1.EndpointTargetSpec spec = 3;
```

temporal/api/cloud/nexus/v1/message.proto
bergundy: I think this should be payload to be future proof.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,85 @@
...
+
+    // Indicates where the endpoint should forward received nexus requests to. 
+    EndpointTargetSpec target_spec = 2;
+
+    // The set of authorization policies for the endpoint. Each request's caller
+    // must match with at least one of the specs to be accepted by the endpoint.
+    // This field is mutable.
+    repeated EndpointAuthorizationSpec authorization_specs = 3;
+
+    // Optional description. Can be in markdown for rendering in the UI
```
mastermanu, line 23: add comment: experimental, this data type may change in the future

temporal/api/cloud/nexus/v1/message.proto
mastermanu: for now, add comment: experimental, this data type may change in the future

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +235,89 @@ message GetRegionResponse {
...
+    // The list of endpoints in ascending id order.
+    repeated temporal.api.cloud.nexus.v1.Endpoint endpoints = 1;
+
+    // The next page's token.
+    string next_page_token = 2;
+}
+
+message GetNexusEndpointRequest {
+    // The id of the nexus endpoint to get.
+    string endpoint_id = 1;
```
bergundy, line 264: I think you can just call this `id`, that's what I did in the operator API.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +235,89 @@ message GetRegionResponse {
...
+message GetNexusEndpointResponse {
+    // The nexus endpoint.
+    temporal.api.cloud.nexus.v1.Endpoint endpoint = 1;
+}
+
+message CreateNexusEndpointRequest {
+    // The spec for the nexus endpoint.
+    temporal.api.cloud.nexus.v1.EndpointSpec spec = 1;
+
+    // The id to use for this async operation - optional.
```
bergundy, line 276: Do we have guidance anywhere explaining why you'd want to specify this? I understand the concept but I'm not sure users do.

temporal/api/cloud/nexus/v1/message.proto
bergundy, line 23:
```suggestion
    // Optional description. Can be in markdown for rendering in the UI.
```

temporal/api/cloud/nexus/v1/message.proto
bergundy: I'm fine with this but I'd like to understand why we can't use what's standard.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,85 @@
...
+    // This field is mutable.
+    string name = 1;
+
+    // Indicates where the endpoint should forward received nexus requests to. 
+    EndpointTargetSpec target_spec = 2;
+
+    // The set of authorization policies for the endpoint. Each request's caller
+    // must match with at least one of the specs to be accepted by the endpoint.
+    // This field is mutable.
+    repeated EndpointAuthorizationSpec authorization_specs = 3;
```
bergundy, line 21: I'm now thinking maybe instead of calling these authorization specs, call them policies.
We can then add rate limiting rules here too.
So I can express constraints like `allow this namespace but limit it to 100 RPS` in the future.

bergundy (commented):
Overall LGTM. I had a few small comments and one that I think we should discuss.

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -125,4 +125,41 @@ service CloudService {
...
+        option (google.api.http) = {
+            post: "/api/v1/nexus/endpoints",
+	    body: "*"
+        };
+    }
+
+    // Update a nexus endpoint
+    rpc UpdateNexusEndpoint(UpdateNexusEndpointRequest) returns (UpdateNexusEndpointResponse) {
+	option (google.api.http) = {
+            post: "/api/v1/nexus/endpoints/{endpoint_id}",
```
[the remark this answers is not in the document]
cretz: I almost missed this because it wasn't marked "ready for review" yet, heh.

We should combine these into one create-or-update, not sure the benefit of separating them. If you combine them you can use this URL, people expect POST to a resource like this to be able to create. If you must keep them separated, since we can't really support `PUT` accurately, I do think it deserves its own operation name.

Is there are good reason here or in operator service why we can't just have one mutation call?

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,85 @@
...
+message EndpointTargetSpec {
+    // The type of target that this endpoint routes Nexus requests to.
+    // Possible values are: cloud_task_queue
+    string target_type = 1;
+
+    // A target spec for routing nexus requests to a specific cloud namespace's taskqueue.
+    CloudTaskQueueSpec cloud_task_queue_spec = 2;
+}
+
+message CloudTaskQueueSpec {
```
bergundy, line 36: In OSS I'm considering calling this `Worker` target.

temporal/api/cloud/nexus/v1/message.proto
mastermanu: added comment

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: talked offline. we are good with keeping this for now

temporal/api/cloud/nexus/v1/message.proto
mastermanu: will follow up with abhinav today just to have a solidified answer / reasoning here.

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: so I changed this to id, but this is now in-consistent with our other APIs related to user and namespace operations (where the request/path takes in user_id or namespace_id). am fine with having this as id, but will check with @anekkanti to be sure

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: Agree that this probably needs to be fleshed out more here: https://docs.temporal.io/ops. The API itself is still in preview.  I have seen other APIs have a section about "idempotency" which describes the usage/semantics around similar fields. We may want something similar her.

cc @anekkanti @jlacefie in case there is some other documentation I am missing here, or to potentially track ensuring docs cover this eventually

temporal/api/cloud/cloudservice/v1/service.proto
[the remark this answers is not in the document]
mastermanu: So few things here:
1) User and Namespace have separate create and update endpoints. We can definitely combine it for Nexus, but then that is inconsistent with the rest of our current Cloud APIs

2) Having distinction between Create/Update has been generally useful for us given that it gives flexibility for the operations to have different auth permissions (and is actually the case for namespaces for instance). Note that from the Cloud Side, create generates an ID that must be passed for Update (same for namespace / user - we take the IDs not the name/email address). We can scrap the ID of the "id" and only rely on the user provided field, but having this decoupled ID seems better in the long-term. 

3) Namespace and User have `POST` as well for update on the resource path itself. If we were to keep these separate, should we update the existing paths on those two types to have an `/update` suffix hanging off of the end as well? I would think so, but just double-checking.

cc @anekkanti to go over this as well

temporal/api/cloud/nexus/v1/message.proto
mastermanu: changed to PolicySpec

temporal/api/cloud/nexus/v1/message.proto
mastermanu: changed it to `CloudWorkerTargetSpec` for now to be more aligned

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,86 @@
...
+    string state = 4;
+
+    // The id of any ongoing async operation that is creating, updating, or deleting the endpoint, if any.
+    string async_operation_id = 5;
+
+    // The date and time when the endpoint was created.
+    google.protobuf.Timestamp created_time = 6;
+
+    // The date and time when the endpoint was last modified.
+    google.protobuf.Timestamp last_modified_time = 7;
```
bergundy, line 83: @mastermanu can we also show the identity that created and modified this endpoint?

mastermanu: i think we can, although we'd probably want to consider adding that across all the objects we have (e.g. users, namespaces, etc).

would probably consider adding that as a separate effort.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +235,89 @@ message GetRegionResponse {
...
+    // The list of endpoints in ascending id order.
+    repeated temporal.api.cloud.nexus.v1.Endpoint endpoints = 1;
+
+    // The next page's token.
+    string next_page_token = 2;
+}
+
+message GetNexusEndpointRequest {
+    // The id of the nexus endpoint to get.
+    string id = 1;
```
cretz, line 264:
```suggestion
    string endpoint_id = 1;
```

Here and elsewhere to match what I see is already done in this repo

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -234,3 +235,89 @@ message GetRegionResponse {
...
+message GetNexusEndpointsRequest {
+    // The requested size of the page to retrieve - optional.
+    // Cannot exceed 1000. Defaults to 100. 
+    int32 page_size = 1;
+
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+
+    // Filter endpoints based on their target specification - optional, treated as an AND if specified. Empty fields within the spec are treated as '*'.
+    temporal.api.cloud.nexus.v1.EndpointTargetSpec spec = 3;
```
cretz, line 248: I think to match what this API does, you should list out specific fields to filter on. I don't think it's consistent with other APIs here to accept an entire spec as a filter.

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -125,4 +125,41 @@ service CloudService {
             get: "/api/v1/cloud/regions/{region}",
         };
     }
+
+       // Gets nexus endpoints
+    rpc GetNexusEndpoints(GetNexusEndpointsRequest) returns (GetNexusEndpointsResponse) {
+	option (google.api.http) = {
+            get: "/api/v1/nexus/endpoints",
```
cretz, line 132:
```suggestion
            get: "/api/v1/cloud/nexus/endpoints",
```

Need to update the prefix to match others in this service (helps routing and may change soon anyways)

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,86 @@
...
+import "google/protobuf/timestamp.proto";
+
+message EndpointSpec {
+    // The name of the endpoint. Must be unique within an account.
+    // The name must match `[a-zA-Z_][a-zA-Z0-9_]*`.
+    // This field is mutable.
+    string name = 1;
+
+    // Indicates where the endpoint should forward received nexus requests to. 
+    EndpointTargetSpec endpoint_target_spec = 2;
```
cretz, line 16:
```suggestion
    EndpointTargetSpec target_spec = 2;
```

Inconsistent in whether you choose to prefix fields in this message with "endpoint_"

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,86 @@
...
+
+    // The set of policies (e.g. authorization) for the endpoint. Each request's caller
+    // must match with at least one of the specs to be accepted by the endpoint.
+    // This field is mutable.
+    repeated EndpointPolicySpec policy_specs = 3;
+
+    // Optional description. Can be in markdown for rendering in the UI.
+    // Note that this is an experimental field that may be deprecated in the future
+    string description = 4;
```
cretz, line 25: We support markdown-based descriptions in API elements? We have some other work trying to define the subset of markdown we're going to support. If we don't have description on namespaces or users or whatever, can we wait until we can consistently provide description on all of these types of resources instead of making Nexus endpoint the first? I support the idea, just think it should be generally applicable.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,86 @@
...
+    string namespace = 1;
+
+    // The task queue on the cloud namespace to route requests to. This field is mutable.
+    string task_queue = 2;
+}
+
+message EndpointPolicySpec {
+    // The policy set on this endpoint authorization spec.
+    // Possible values are: allowed_cloud_namespace
+    string policy_type = 1;
```
cretz, line 48: Is this necessary? Can the presence of the next field be enough to tell you what the policy type is?

temporal/api/cloud/nexus/v1/message.proto
cretz: Same, seems unnecessary that you can't use the presence of the next field to determine type

temporal/api/cloud/cloudservice/v1/service.proto
mastermanu: whoops, fixed

temporal/api/cloud/nexus/v1/message.proto
mastermanu: removed endpoint_ suffix

temporal/api/cloud/nexus/v1/message.proto
mastermanu: this is definitely open/experimental. we dont support description on any other resources (except for service accounts, which is also `string description`, and thats not released yet). Am fine removing this for now and adding it later. It's also why it was marked as "experimental" via a comment. @bergundy - any thoughts here?

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: sure, can replace with flat fields

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: @rachfop since he is working on docs

temporal/api/cloud/cloudservice/v1/service.proto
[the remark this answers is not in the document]
mastermanu: Synced with @anekkanti offline - we will add "update" suffix to the other update operation paths here and keep CREATE/UPDATE as separate operations for the reasons stated above.

temporal/api/cloud/nexus/v1/message.proto
mastermanu: Spoke with @anekkanti offline. as far as we can tell there are four options here:

1) use oneoff. I think we just decided not to use that because there were concerns about around how this manifests in other languages, and even the generated go code to read the field seems janky: https://protobuf.dev/reference/go/go-generated/#oneof. We were also worried that adding new types in the future would complicate versioning. Maybe both of these worries are unfounded. This is the first time we are adding something like this to the Cloud API, so worth laying the groundwork for this now. Curious if @cretz or @bergundy have thoughts on "oneoff" and cross-platform/language usage and any issues around that.

2) Don't use `oneoff`, but don't have the `target_type` field, just check that exactly one field is assigned a value. No reason we cannot do this. If I wanted to do a Get call for all endpoints of a specific target type, we would probably still need the string on that `GetNexusEndpoints` get request, but probably not a big deal. 

3) Use `target_type` as an `enum` (I think there was a previous discussion about not using enums, we can re-open that if need be).

4) Use `target_type` as string (as is now). It's a bit more explicit, but agree its unnecessary. We'd probably still include it as an optional filter on GetNexusEndpoints() [but it can be added later down the line since there is only one type for now anyway]

Leaning towards option 2 (just get rid of `target_type` field). Any concerns with that?

cretz: I too lean towards option 2. Unrelated to this specifically, but I sure would like to re-open that never-use-proto-enum decision that was made long ago on this repo. I disagree with it.

temporal/api/cloud/cloudservice/v1/service.proto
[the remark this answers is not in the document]
cretz: Now that users can't provide their own ID, I don't mind if you leave it as is now FWIW. My primary concern was when they could provide their own ID, then the ID-based URL should be create-or-update.

temporal/api/cloud/cloudservice/v1/request_response.proto
cretz: We need to be consistent with ourselves, I updated that thread too

temporal/api/cloud/nexus/v1/message.proto
mastermanu: ok cool. removed it completely for now. we have only one type so we can figure out if we want to add target_type later for the search/get

temporal/api/cloud/cloudservice/v1/request_response.proto
mastermanu: changed it back

temporal/api/cloud/cloudservice/v1/service.proto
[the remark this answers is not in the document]
mastermanu: reverted

temporal/api/cloud/nexus/v1/message.proto
mastermanu: removed for now. can add it in a separate PR after more discussion

temporal/api/cloud/nexus/v1/message.proto
mastermanu: removed

cretz (approved):

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -235,6 +236,29 @@ message GetRegionResponse {
...
+
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+
+    // optional, treated as an AND if specified
+    string target_namespace_id = 3;
+
+    // optional, treated as an AND if specified
+    string target_task_queue = 4;
+
```
bergundy, line 252: We can live without this for now and filter in the UI, I'd consider removing until there's demand.

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,74 @@
...
+    // This field is mutable.
+    repeated EndpointPolicySpec policy_specs = 3;
+}
+
+message EndpointTargetSpec {
+    // A target spec for routing nexus requests to a specific cloud namespace worker.
+    CloudWorkerTargetSpec cloud_worker_target_spec = 1;
+}
+
+message CloudWorkerTargetSpec {
```
bergundy, line 29: nit: I'd just call this `Worker`, it's not a cloud worker.

bergundy (approved):

temporal/api/cloud/nexus/v1/message.proto
bergundy: I missed this conversation.

Our designs all include description in markdown, we have to keep this field.
I have requested that we make it a `Payload` so eventually, when we support account level codecs in the UI we can encrypt them.

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -179,6 +179,42 @@ service CloudService {
...
+	          body: "*"
+        };
+    }
+
+    // Delete a nexus endpoint
+    rpc DeleteNexusEndpoint(DeleteNexusEndpointRequest) returns (DeleteNexusEndpointResponse) {
+	      option (google.api.http) = {
+            delete: "/cloud/nexus/endpoints/{endpoint_id}",
+        };
+    }
```
cretz, line 217:
```suggestion
    }

```

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -179,6 +179,42 @@ service CloudService {
         };
     }
 
+    // Gets nexus endpoints
+    rpc GetNexusEndpoints(GetNexusEndpointsRequest) returns (GetNexusEndpointsResponse) {
+	      option (google.api.http) = {
```
cretz, line 184:
```suggestion
        option (google.api.http) = {
```

Spacing off elsewhere for this too. Make sure you are not using tabs.

cretz (approved):

temporal/api/cloud/cloudservice/v1/service.proto
bergundy: Is it time to add a formatter?

cretz: I think that can be discussed on the primary `api` repo and can apply to this if/when it moves there

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,75 @@
...
+    // The next update operation must include this version.
+    string resource_version = 2;
+
+    // The endpoint specification.
+    EndpointSpec spec = 3;
+
+    // The current state of the endpoint.
+    // Possible values: activating, activationfailed, active, updating, updatefailed, deleting, deletefailed, deleted
+    // For any failed state, reach out to Temporal Cloud support for remediation.
+    string state = 4;
```
anekkanti, line 65: Lets move to using an enum here. 
@nikki-dag

temporal/api/cloud/nexus/v1/message.proto
```diff
@@ -0,0 +1,75 @@
...
+    // This field is mutable.
+    repeated EndpointPolicySpec policy_specs = 3;
+
+    // The markdown description of the endpoint - optional.
+    string description = 4;
+}
+
+message EndpointTargetSpec {
+    // A target spec for routing nexus requests to a specific cloud namespace worker.
+    WorkerTargetSpec worker_target_spec = 1;
```
bergundy, line 29: @nikki-dag this would be an enum (with a single variant ATM)

temporal/api/cloud/nexus/v1/message.proto
anekkanti: We should reuse this: https://github.com/temporalio/api-cloud/pull/36/files#diff-b38afecfa7fefa137fb5dbe446fd43bd6603676748e6f061fcf1f03d2e386706R13

PR #24, opened 2024-05-07
Proposes visibility field for namespace

mastermanu:


anekkanti:
@mastermanu is this change still relevant? If not can we close this PR.

PR #29, opened 2024-05-14
Update Multi-Region Namespace APIs

yux0:
## What was changed
Update Multi-Region Namespace APIs

## Why?
Update Multi-Region Namespace APIs

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -223,10 +223,8 @@ message FailoverNamespaceRegionRequest {
     // The id of the region to failover to.
     // Must be a region that the namespace is currently available in.
     string region = 2;
-    // Skip graceful failover and only start forceful failover.
-    bool skip_graceful_failover = 3;
     // The id to use for this async operation - optional.
-    string async_operation_id = 4;
+    string async_operation_id = 3;
 }
 
 message FailoverNamespaceRegionResponse {
@@ -235,9 +233,9 @@ message FailoverNamespaceRegionResponse {
 }
 
 message AddNamespaceRegionRequest {
-    // The namespace to add region to.
+    // The namespace to add the region to.
     string namespace = 1;
-    // The id of the region to add to the namespace.
+    // The id of the standby region to add to the namespace.
     string region = 2;
     // The version of the namespace for which this add region operation is intended for.
     // The latest version can be found in the GetNamespace operation response.
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -217,6 +217,40 @@ message DeleteNamespaceResponse {
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
+message FailoverNamespaceRegionRequest {
+    // The namespace to failover.
+    string namespace = 1;
+    // The id of the region to failover to.
+    // Must be a region that the namespace is currently available in.
+    string region = 2;
+    // Skip graceful failover and only start forceful failover.
+    bool skip_graceful_failover = 3;
+    // The id to use for this async operation - optional.
+    string async_operation_id = 4;
+}
+
+message FailoverNamespaceRegionResponse {
+    // The async operation.
+    temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
+}
+
+message AddNamespaceRegionRequest {
+    // The namespace to add region to.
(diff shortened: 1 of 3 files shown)


temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -112,6 +112,22 @@ service CloudService {
         };
     }
 
+    // Failover a multi-region namespace
+    rpc FailoverNamespaceRegion (FailoverNamespaceRegionRequest) returns (FailoverNamespaceRegionResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/failover",
```
```suggestion
            post: "/api/v1/cloud/namespaces/{namespace}/failover-region",
```

Need different prefix, and I think we should add region to the URL instead of generic "failover" (unless there is a good case for calling this RPC `FailoverNamespace` instead of `FailoverNamespaceRegion`)

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -112,6 +112,22 @@ service CloudService {
...
+            body: "*"
+        };
+    }
+
+    // Add a new region to a namespace
+    rpc AddNamespaceRegion (AddNamespaceRegionRequest) returns (AddNamespaceRegionResponse) {
+        option (google.api.http) = {
+            post: "/api/v1/namespaces/{namespace}/add-region",
```
cretz, line 126:
```suggestion
            post: "/api/v1/cloud/namespaces/{namespace}/add-region",
```

cretz (approved):

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -217,6 +217,40 @@ message DeleteNamespaceResponse {
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
+message FailoverNamespaceRegionRequest {
+    // The namespace to failover.
+    string namespace = 1;
+    // The id of the region to failover to.
+    // Must be a region that the namespace is currently available in.
+    string region = 2;
+    // Skip graceful failover and only start forceful failover.
```
mastermanu, line 226: this can essentially result in data-loss / zombies right? Do we need to have a stronger comment here cautioning use of this?

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -217,6 +217,40 @@ message DeleteNamespaceResponse {
...
+    string async_operation_id = 4;
+}
+
+message FailoverNamespaceRegionResponse {
+    // The async operation.
+    temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
+}
+
+message AddNamespaceRegionRequest {
+    // The namespace to add region to.
```
mastermanu, line 238: "The namespace to add the region to"

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -217,6 +217,40 @@ message DeleteNamespaceResponse {
...
+    string namespace = 1;
+    // The id of the region to add to the namespace.
```
mastermanu, line 240: Do we need any clarification comment here that this is secondary/passive region?

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -130,4 +130,16 @@ message Namespace {
     // The date and time when the namespace was last modified.
     // Will not be set if the namespace has never been modified.
     google.protobuf.Timestamp last_modified_time = 11;
+    // The status of each region where the namespace is available.
+    // The id of the region is the key and the status is the value of the map.
+    map<string, NamespaceRegionStatus> region_status = 12;
+}
+
+message NamespaceRegionStatus {
+    // The current state of the namespace region.
```
mastermanu, line 139: [Nit] might want to reword this a bit as this could be interpreted as "the state of the whole region the namespace is in"
Maybe, the current state of the namespace in the given region or something would be better?

mastermanu (approved):

temporal/api/cloud/cloudservice/v1/request_response.proto
yux0: This is a good callout. I will remove this one for now and if necessary, we can expose it later.

temporal/api/cloud/namespace/v1/message.proto
yux0: This is the state of a region where the namespace is in.

PR #33, opened 2024-05-29
[Breaking] Reorg the user-group apis with a explicit google group spec field.

anekkanti:
This change is backward incompatible. But we control who and how its being used. Should be safe to merge.

### DIFF
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -360,13 +360,13 @@ message GetUserGroupsRequest {
     string namespace = 3;
     // Filter groups by the display name - optional.
     string display_name = 4;
+    // Filter groups by the google group specification - optional.
+    GoogleGroupFilter google_group = 5;
 
     message GoogleGroupFilter {
         // Filter groups by the google group email - optional.
         string email_address = 1;
     }
-    // Filter groups by the google group specification - optional.
-    GoogleGroupFilter google_group = 5;
 }
 
 message GetUserGroupsResponse {
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -349,7 +349,6 @@ message DeleteApiKeyResponse {
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
-
 message GetUserGroupsRequest {
     // The requested size of the page to retrieve - optional.
     // Cannot exceed 1000. Defaults to 100.
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -359,12 +359,9 @@ message GetUserGroupsRequest {
     string namespace = 3;
     // Filter groups by the display name - optional.
     string display_name = 4;
-    // Filter groups by the group type - optional.
-    // Possible values: google-group
-    string group_type = 5;
     // Filter groups by the google group email - optional.
-    // Only applicable when group_type filter is set to google group-type.
-    string google_group_email_address = 6;
+    // Only applies to google groups.
+    string google_group_email_address = 5;
 }
 
 message GetUserGroupsResponse {
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -349,10 +349,6 @@ message DeleteApiKeyResponse {
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
-message GoogleGroupGetUserGroupsFilter {
-    // Filter groups by the google group email - optional.
-    string email_address = 1;
-}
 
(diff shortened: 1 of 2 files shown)


temporal/api/cloud/identity/v1/message.proto
```diff
@@ -70,12 +70,24 @@ message User {
...
-    string name = 1;
-    // The access assigned to the group
+    // The display name of the group.
+    string display_name = 1;
+    // The access assigned to the group.
     Access access = 2;
+    // The type of the group.
+    // Currently, the only supported type is 'google-group'.
+    // The type is immutable. Once set during creation, it cannot be changed.
+    string type = 3;
```
codemonkeycxy, line 87: we can chat about this today. last time @mattkim suggested not exposing this field yet. maybe the presence of GoogleGroupSpec would indicate the group type implicitly? I will take a look at the implementation to see the necessity of exposing this field now. but just wanna leave a reminder for myself to discuss

codemonkeycxy (approved):

temporal/api/cloud/identity/v1/message.proto
```diff
@@ -70,12 +70,24 @@ message User {
...
+    // The specification of the google group.
+    // This field is required when the group type is 'google-group'.
+    GoogleGroupSpec google_group = 4;
```
cretz, line 90: Can't the presence of `google_group` be enough to not need `type`? Would you consider a `oneof`? Same for the filter on the request.

mattkim (approved):

temporal/api/cloud/identity/v1/message.proto
anekkanti: I think I will sight the discussion here:
https://github.com/temporalio/api-cloud/pull/23#discussion_r1593105415
We decided not to use one-off but rely on presence or absence of fields instead of a explicit type.
Let me remove the type field.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -357,8 +357,14 @@ message GetUserGroupsRequest {
...
-    // Filter groups by their name - optional.
-    string group_name = 4;
+    // Filter groups by the display name - optional.
+    string display_name = 4;
+    // Filter groups by the group type - optional.
+    // Possible values: google-group
+    string group_type = 5;
+    // Filter groups by the google group email - optional.
+    // Only applicable when group_type filter is set to google group-type.
+    string google_group_email_address = 6;
```
cretz: Can you do the same thing here as you did in the spec? Specifically can you have a `GoogleGroupFilter` that can contain filterable fields and even just its empty presence is like `group_type`? My concern is that you're going to add a new top-level field for every field inside every spec type, and they only work if they are set as the same time as something else which is confusing.

anekkanti: Just updated and removed the type field.

> and even just its empty presence is like group_type
Proto3 does not distinguish between empty and unset scalars. 
https://github.com/protocolbuffers/protobuf/issues/1606#issuecomment-281832148
I don't think we can implement a type filter with just a googleGroupFilter.

cretz: > Proto3 does not distinguish between empty and unset scalars.

That is unrelated to my concern now about top-level fields. Are you going to add a new top-level field for every field you want to filter on each group type? My suggestion is just making a a child object for all google group filter fields (and another for the next one) instead of moving them all top level. Same reason you have `UserGroupSpec.google_group.email_address` instead of `UserGroupSpec.google_group_email_address, you should have `GetUserGroupsRequest.google_group.email_address` instead of `GetUserGroupsRequest.google_group_email_address`.

anekkanti: Got it. Let me do that.

anekkanti: Got it, makes sense. Added a child object as suggested.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -349,6 +349,11 @@ message DeleteApiKeyResponse {
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
 
+message GoogleGroupGetUserGroupsFilter {
```
cretz, line 352: Not the biggest fan of this name (I'd just have made a nested `GoogleGroupFilter` in `GetUserGroupsRequest`) but not that big of a deal

cretz (approved):

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -357,8 +358,15 @@ message GetUserGroupsRequest {
     string page_token = 2;
     // Filter groups by the namespace they have access to - optional.
     string namespace = 3;
-    // Filter groups by their name - optional.
-    string group_name = 4;
+    // Filter groups by the display name - optional.
+    string display_name = 4;
+
+    message GoogleGroupFilter {
```
cretz, line 364: Personally I think it's best to put all nested messages at the bottom of the parent, though some proto authors put them at the top, but I don't think it's very common to intersperse between fields.

PR #34, opened 2024-06-14
add more comments around region ids

anekkanti:


temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -236,6 +236,7 @@ message AddNamespaceRegionRequest {
     // The namespace to add the region to.
     string namespace = 1;
     // The id of the standby region to add to the namespace.
+    // The GetRegions API can be used to get the list of valid region ids.
```
jumbogo, line 239: nit Do you want to add some explanation on the provider prefix?

jumbogo (approved):

temporal/api/cloud/cloudservice/v1/request_response.proto
anekkanti: Don't know if talking about the structure of the id is a good idea.
Added an example, hopefully that will help with making it clearer.

PR #35, opened 2024-07-18
tmp

shivam-ajmera:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal/api/cloud/auditlog/v1/message.proto
+++ b/temporal/api/cloud/auditlog/v1/message.proto
@@ -1,0 +1,107 @@
+syntax = "proto3";
+
+package temporal.api.cloud.auditlog.v1;
+
+option go_package = "go.temporal.io/api/cloud/auditlog/v1;auditlog";
+option java_package = "io.temporal.api.cloud.auditlog.v1";
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Cloud::AuditLog::V1";
+option csharp_namespace = "Temporalio.Api.Cloud.AuditLog.V1";
+
+import "google/protobuf/timestamp.proto";
+
+// LogRecord represents an audit log entry from Temporal, structured for easy parsing and analysis.
+message LogRecord {
+  // Time when the log was emitted from the source
+  google.protobuf.Timestamp emit_time = 1;
+
+  // Level of the log, i.e. info, warn, error etc
+  LogLevel level = 2;
+
+  // Operator email address
+  string user_email = 3;
+
+  // Operator IP address or server name
+  string caller_ip_address = 4;
+
+  // Operation performed
+  string operation = 5;
+
+  // Detailed information about the operation.
+  OperationDetail details = 6;
+
+  // Operation status from API call
+  string status = 7;
+
+  // Admin or System
+  LogCategory category = 8;
+
+  // Version of the log entry
+  int32 version = 9;
+
+  // Unique ID for the log record.
+  string log_id = 10;
+}
+
+// OperationDetails includes potential more detailed operation logs extracted from the different fields in the
+// Temporal API logs or third party logs, e.g. auth0
+message OperationDetail {
+  // The namespace the operation was performed on
+  string namespace = 1;
+
+  // The users the operation was performed on
+  repeated string target_users = 2;
+
+  // Role information if any role is processed
+  repeated string roles = 3;
+
+  // Fingerprints if client ca is included
+  repeated string client_ca_fingerprints = 4;
+
+  // Search attributes if the search attributes were processed
+  SearchAttributeUpdate search_attribute_update = 5;
+
+  // Additional message if there is any
+  string additional_message = 6;
+
+  // Namespace's region if available
+  string region = 7;
+
+  // If account features were updated i.e. external metrics, audit logging sink etc.
+  repeated string account_features = 8;
+
+  // Third party log details eg: auth0
+  ThirdPartyLogDetail third_party_details = 9;
+}
+
+message SearchAttributeUpdate {
+  string existing_attribute = 1;
+
+  string new_attribute = 2;
+}
+
+message ThirdPartyLogDetail {
+  string id = 1;
+
+  // the third party source, e.g. auth0
+  string source = 2;
+
+  // the raw log message from the third party app, e.g. the whole auth0 event log
+  string raw_message = 3;
+}
+
+enum LogCategory {
+  LOG_CATEGORY_UNSPECIFIED = 0;
+  LOG_CATEGORY_ADMIN = 1;
+  LOG_CATEGORY_SYSTEM = 2;
+}
+
+enum LogLevel {
+  LOG_LEVEL_UNSPECIFIED = 0;
+  LOG_LEVEL_INFO = 1;
+  LOG_LEVEL_DEBUG = 2;
+  LOG_LEVEL_WARN = 3;
+  LOG_LEVEL_ERROR = 4;
+  LOG_LEVEL_FATAL = 5;
+}
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -9,6 +9,7 @@ option java_outer_classname = "RequestResponseProto";
 option ruby_package = "Temporalio::Api::Cloud::CloudService::V1";
 option csharp_namespace = "Temporalio.Api.Cloud.CloudService.V1";
 
+import "temporal/api/cloud/auditlog/v1/message.proto";
 import "temporal/api/cloud/operation/v1/message.proto";
 import "temporal/api/cloud/identity/v1/message.proto";
 import "temporal/api/cloud/namespace/v1/message.proto";
@@ -524,4 +525,25 @@ message DeleteServiceAccountRequest {
 message DeleteServiceAccountResponse {
     // The async operation.
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
+}
+
+message GetAuditLogsRequest {
+    // The requested size of the page to retrieve - optional.
+    // Cannot exceed 1000. Defaults to 100.
+    int32 page_size = 1;
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+    // Filter for logs inserted at a time >= this.
+    google.protobuf.Timestamp start_time_inclusive = 2;
+    // Filter for logs inserted at a time < this. - optional.
+    google.protobuf.Timestamp end_time_exclusive = 3;
+    // Filter logs by a user's email address - optional.
+    string user_email_substring = 4;
+}
+
+message GetAuditLogsResponse {
+    // The list of audit logs ordered by inserted time, emit time, log_id
+    repeated temporal.api.cloud.auditlog.v1.LogRecord groups = 1;
+    // The next page's token.
+    string next_page_token = 2;
 }


temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+enum LogCategory {
+  LOG_CATEGORY_UNSPECIFIED = 0;
+  LOG_CATEGORY_ADMIN = 1;
+  LOG_CATEGORY_SYSTEM = 2;
+}
+
+enum LogLevel {
+  LOG_LEVEL_UNSPECIFIED = 0;
+  LOG_LEVEL_INFO = 1;
+  LOG_LEVEL_DEBUG = 2;
```
cretz, line 103: I would expect debug to be above this (also I am unfamiliar with audit logging, but I would usually recommend a TRACE even above debug for future compat).

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+  // Time when the log was emitted from the source
+  google.protobuf.Timestamp emit_time = 1;
+
+  // Level of the log, i.e. info, warn, error etc
+  LogLevel level = 2;
+
+  // Operator email address
+  string user_email = 3;
+
+  // Operator IP address or server name
```
cretz, line 25: If it can be either, field should not be suffixed with `ip_address`. Maybe a more generic `host` or `address`?

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+  string log_id = 10;
+}
+
+// OperationDetails includes potential more detailed operation logs extracted from the different fields in the
+// Temporal API logs or third party logs, e.g. auth0
+message OperationDetail {
+  // The namespace the operation was performed on
+  string namespace = 1;
+
+  // The users the operation was performed on
```
cretz, line 53: User what, IDs? Emails?

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+  repeated string target_users = 2;
+
+  // Role information if any role is processed
```
cretz, line 56: What is mean by "information"? Is this role names?

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+  repeated string roles = 3;
+
+  // Fingerprints if client ca is included
+  repeated string client_ca_fingerprints = 4;
+
+  // Search attributes if the search attributes were processed
```
cretz, line 62: Comment implies plural but value is singular

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+  string user_email = 3;
+
+  // Operator IP address or server name
+  string caller_ip_address = 4;
+
+  // Operation performed
+  string operation = 5;
+
+  // Detailed information about the operation.
+  OperationDetail details = 6;
```
cretz, line 32: Either call messages `OperationDetails` or field name `detail`. No need to mix singular and plural in this case IMO.

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+  string status = 7;
+
+  // Admin or System
+  LogCategory category = 8;
+
+  // Version of the log entry
+  int32 version = 9;
+
+  // Unique ID for the log record.
+  string log_id = 10;
```
cretz, line 44: For other things that have an ID, like `User`, we don't prefix the type of ID if it applies to the proto message as a whole. May be best as just `id`.

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+  // Detailed information about the operation.
+  OperationDetail details = 6;
+
+  // Operation status from API call
+  string status = 7;
+
+  // Admin or System
+  LogCategory category = 8;
+
+  // Version of the log entry
```
cretz, line 40: Just for my own curiosity, what does "version" mean here?

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+option java_package = "io.temporal.api.cloud.auditlog.v1";
+option java_multiple_files = true;
+option java_outer_classname = "MessageProto";
+option ruby_package = "Temporalio::Api::Cloud::AuditLog::V1";
+option csharp_namespace = "Temporalio.Api.Cloud.AuditLog.V1";
+
+import "google/protobuf/timestamp.proto";
+
+// LogRecord represents an audit log entry from Temporal, structured for easy parsing and analysis.
+message LogRecord {
```
cretz, line 15: Are all log records a result of an operation? Is there a need to abstract this, e.g. put everything related to an operation in a separate message/field and leave log record generic for other forms of log records later? Having `operation` and `status` here while also having operation details as separate is a bit offputting.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -525,3 +528,24 @@ message DeleteServiceAccountResponse {
     // The async operation.
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
+
+message GetAuditLogsRequest {
```
cretz, line 532: Is the service entry missing for this req/resp RPC?

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -525,3 +528,24 @@ message DeleteServiceAccountResponse {
...
+    // Cannot exceed 1000. Defaults to 100.
+    int32 page_size = 1;
+    // The page token if this is continuing from another response - optional.
+    string page_token = 2;
+    // Filter for logs inserted at a time >= this.
+    google.protobuf.Timestamp start_time_inclusive = 3;
+    // Filter for logs inserted at a time < this. - optional.
+    google.protobuf.Timestamp end_time_exclusive = 4;
+    // Filter logs by a user's email address - optional.
+    string user_email_substring = 5;
```
cretz, line 543: Are other filters in this file for email address when listing also substring or equality match? Is substring important and/or does it put us in a bind later if we have this in some form of database since substrings are hard to have indexes on?

temporal/api/cloud/auditlog/v1/message.proto
shivam-ajmera: Yes all records should be the result of an operation. Status is technically one of `OK` & `ERROR` (could mention this but it's in docs already fwiw)

temporal/api/cloud/auditlog/v1/message.proto
shivam-ajmera: Currently, we're migrating from Audit Logging v0 -> v1 (BTS change of our own internal system). Used to differentiate those for now.

temporal/api/cloud/auditlog/v1/message.proto
shivam-ajmera: Yup, updated

temporal/api/cloud/auditlog/v1/message.proto
shivam-ajmera: Emails, good call - updated it

temporal/api/cloud/auditlog/v1/message.proto
shivam-ajmera: Updated

temporal/api/cloud/cloudservice/v1/request_response.proto
shivam-ajmera: Yup just added

temporal/api/cloud/auditlog/v1/message.proto
shivam-ajmera: Uh I'm actually not entirely sure of this one just copied over the proto. We use x-forwarded-for to populate this field, I think that's IP only?

temporal/api/cloud/auditlog/v1/message.proto
jumbogo: I think we should keep the field name here, and maybe rename the message.  Note this format is already [published](https://docs.temporal.io/cloud/audit-logging#audit-log-format) for our customers to refer to.  When we change the fields we need to update the public facing document.  We can schedule an update when we release the endpoint and update the format after next week's v2 rollout.

cretz: Changing the message name here is fine.

I do think this review should have the freedom to not have to match a published format for this beta cloud API. If it's just a doc update, no biggie IMO.

temporal/api/cloud/auditlog/v1/message.proto
cretz: Hrmm, I fear that baking in that forced assumption may harm reusability of this log record for other auditable things. I would encourage anything operation specific to be in its own field. It is confusing that half operation-specific things are top-level and half in details.

temporal/api/cloud/auditlog/v1/message.proto
rohitgup14: Agree, we should use `host` (preferred) or `address`. @shivam-ajmera we should have ability to do some transformation of field names before exposing it via API

temporal/api/cloud/auditlog/v1/message.proto
cretz: Based on internal discussion, we are not allowed to change structure in this review.

cretz, line 15: Can you update fields here (and in child fields) saying which fields may be zero/empty/unset/optional from the server? Thanks!

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+// Temporal API logs or third party logs, e.g. auth0
+message OperationDetail {
+  // The namespace the operation was performed on
+  string namespace = 1;
+
+  // The email addresses of the users on whom the operation was performed
+  repeated string target_users = 2;
+
+  // Names of any roles processed
+  repeated string roles = 3;
```
anekkanti, line 57: Is this role ids? Whats an example of operation with role ids?
We do not expose roles or role ids in the new apis..

temporal/api/cloud/auditlog/v1/message.proto
```diff
@@ -0,0 +1,107 @@
...
+  // Version of the log entry
+  int32 version = 9;
+
+  // Unique ID for the log record.
+  string log_id = 10;
+}
+
+// OperationDetails includes potential more detailed operation logs extracted from the different fields in the
+// Temporal API logs or third party logs, e.g. auth0
+message OperationDetail {
```
anekkanti, line 49: Do we have flexibility to structure this better. 
My worry is this message will become a dumping ground of every information of the about the operation.
From the top of my head can we have a NamespaceDetails, UserDetails, AccountDetails, ServiceAccountDetails, ApikeyDetails, UserGroupDetails, etc..

cretz: I was told we cannot change this structure at this time. If we can, I'd like to request all operation-specific things be in the operation area (including operation name and status) instead of just some

temporal/api/cloud/auditlog/v1/message.proto
shivam-ajmera: Done, thanks!

temporal/api/cloud/auditlog/v1/message.proto
shivam-ajmera: These are roles associated with users, and yes they're part of the older apis which are supported by audit logging. Will be referred to accesses associated with identities in the newer apis

shivam-ajmera:
Resolved comments based on internal discussions as updating schema is not part of this release.

cretz:
Ah, I think my approval is based on the ability to review the fields and their names, but if not, can bypass the need for my approval here

PR #36, opened 2024-07-24
[Breaking] Replace strings with enums where ever applicable.

anekkanti:
This a breaking change, but is backward wire compatible.
Meaning the existing clients will keep working when this change is deployed on the server
But the consumers will see compile time errors when they pull the change.

String fields being replaced with enums:

- Namespace::state
- User::state
- ServivceAccount::state
- ApiKey:state
- AsyncOperation::state
- NamespaceRegionStatus::state
- AccountAccess::role
- NamespaceAccess::permission
- ApiKeySpec::owner_type
- NamespaceSpec::custom_search_attributes

Replace string (base64 encoded pem certs) ca certs with a bytes field (pem certs):
- NamespaceSpec::mtls_auth::accepted_client_ca


cretz (approved):

temporal/api/cloud/identity/v1/message.proto
```diff
@@ -9,22 +9,52 @@ option java_outer_classname = "MessageProto";
...
     // developer - gives access to create namespaces on the account
     // read - gives read only access to the account
-    string role = 1;
+    // Deprecated: Use role field instead.
+    string role_deprecated = 1 [deprecated = true];
+    // The role on the account.
+    // temporal:enums:replaces=role_deprecated
+    Role role = 2;
+
+    enum Role {
```
mattkim, line 26: Does defining it this way prevent us from referencing these role enums anywhere else?

temporal/api/cloud/identity/v1/message.proto
```diff
@@ -9,22 +9,52 @@ option java_outer_classname = "MessageProto";
...
     // read - gives read only access to the namespace configuration and workflows within the namespace
-    string permission = 1;
+    // Deprecated: Use permission field instead.
+    string permission_deprecated = 1 [deprecated = true];
+
+    // The permission to the namespace.
+    // temporal:enums:replaces=permission_deprecated
+    Permission permission = 2;
+
+    enum Permission {
```
mattkim, line 46: same here

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -158,7 +164,20 @@ message NamespaceRegionStatus {
...
-    string state = 1;
+    // Deprecated: Use state instead.
+    string state_deprecated = 1 [deprecated = true];
+    // The current state of the namespace region.
+    // temporal:enums:replaces=state_deprecated
+    State state = 3;
     // The id of the async operation that is making changes to where the namespace is available, if any.
     string async_operation_id = 2;
+
+    enum State {
```
mattkim, line 175: same here

buf.yaml
```diff
@@ -3,7 +3,7 @@ deps:
   - buf.build/googleapis/googleapis
 breaking:
   use:
-    - FILE
+    - WIRE
```
mattkim, line 6: just curious what does this do?

mattkim (approved):

temporal/api/cloud/identity/v1/message.proto
anekkanti: Not really, you can still reference it from other places. The enum will be scoped into the message. 
For e.g. when you refer the role you will have to refer it like this: 
```
    temporal.api.cloud.identity.v1.AccountAccess.Role role = 6;
```

buf.yaml
anekkanti: Ohh, this is temporary, since its a backward incompatible change, this fails the build. 
The plan is to reset this to FILE when this PR is merged.

PR #41, opened 2024-09-24
Add Usage API

shivam-ajmera:
## What was changed
Add a Usage API

## Why?
To track usage data across namespaces

## Checklist
- [x] Generated protos with make proto.

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/temporal/api/cloud/cloudservice/v1/request_response.proto
+++ b/temporal/api/cloud/cloudservice/v1/request_response.proto
@@ -9,11 +9,14 @@ option java_outer_classname = "RequestResponseProto";
 option ruby_package = "Temporalio::Api::Cloud::CloudService::V1";
 option csharp_namespace = "Temporalio.Api.Cloud.CloudService.V1";
 
+import "google/protobuf/timestamp.proto";
+
 import "temporal/api/cloud/operation/v1/message.proto";
 import "temporal/api/cloud/identity/v1/message.proto";
 import "temporal/api/cloud/namespace/v1/message.proto";
 import "temporal/api/cloud/nexus/v1/message.proto";
 import "temporal/api/cloud/region/v1/message.proto";
+import "temporal/api/cloud/usage/v1/message.proto";
 
 message GetUsersRequest {
     // The requested size of the page to retrieve - optional.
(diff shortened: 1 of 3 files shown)


temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -619,3 +622,33 @@ message DeleteServiceAccountResponse {
     // The async operation.
     temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
 }
+
+message GetUsageRequest {
+    // Filter for UTC time >= - optional.
+    // Defaults to: start of the current month.
+    // Must be: within the last 90 days from the current date.
+    // Must be: midnight UTC time.
```
robzienert, line 630: Meaning, if any timestamp that is not midnight UTC time--regardless of the date--validation will fail?

temporal/api/cloud/usage/v1/message.proto
```diff
@@ -0,0 +1,57 @@
...
+    // End of UTC day for now (exclusive)
+    google.protobuf.Timestamp end_time = 2;
+    // Records grouped by namespace
+    repeated RecordGroup record_groups = 3;
+    // True if data for given time window is not fully available yet (e.g. delays)
+    bool incomplete = 4;
+}
+
+message RecordGroup {
+    repeated GroupBy group_bys = 1;
```
robzienert, line 26: 🤔 Curious about this usage. Is it intended that `GroupByKey` would become a logical `AND` if multiple are provided?

temporal/api/cloud/cloudservice/v1/request_response.proto
shivam-ajmera: That is correct, currently it's a strict requirement since we only support `Day` granularity

temporal/api/cloud/usage/v1/message.proto
shivam-ajmera: Good question, that is correct. Currently it defaults to (and allows) only 1 key i.e. grouping by namespace. But multiple would mean logical `AND`

cretz (approved):

temporal/api/cloud/usage/v1/message.proto
```diff
@@ -19,6 +19,7 @@ message Summary {
     // Records grouped by namespace
     repeated RecordGroup record_groups = 3;
     // True if data for given time window is not fully available yet (e.g. delays)
+    // When true, records could be added in the future (until false)
```
mastermanu, line 22: "records could be added or updated for the given start_time and end_time" or something like that

mastermanu (approved):

temporal/api/cloud/usage/v1/message.proto
shivam-ajmera: Updated: `When true, records for the given time range could still be added/updated in the future (until false)`

PR #42, opened 2024-10-02
Add Export Sink Api

alice-yin:
- **add export api**
- **add export sink api**

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


temporal/api/cloud/sink/v1/message.proto
```diff
@@ -0,0 +1,109 @@
...
+    // The receiving sink type for export.
+    ExportDestinationType destination_type = 3;
+    
+    // The S3 configuration details when destination_type is S3.
+    S3Spec s3_sink = 4;
+
+    // The GCS configuration details when destination_type is GCS.
+    GCSSpec gcs_sink = 5;
+
+    enum ExportDestinationType {
```
anekkanti, line 31: Don't think we need this, look at my comment above.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -652,3 +653,95 @@ message GetUsageResponse {
     // The next page's token.
     string next_page_token = 2;
 }
+
+message CreateExportSinkRequest {
```
anekkanti, line 657: Can we add a `Namespace` qualifier to all the request, responses, apis to make it clear that these are namespace export sinks. 

Essentially this becomes `CreateNamespaceExportSinkRequest`

temporal/api/cloud/sink/v1/message.proto
```diff
@@ -0,0 +1,109 @@
...
+        EXPORT_DESTINATION_TYPE_UNSPECIFIED = 0;
+        EXPORT_DESTINATION_TYPE_S3 = 1;
+        EXPORT_DESTINATION_TYPE_GCS = 2;
+    }
+}
+
+
+message ExportSink {
```
anekkanti, line 39: Will this sink be used for anything other then for namespace exports? If no, we should move it to the namespace package, i think.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -652,3 +653,95 @@ message GetUsageResponse {
...
+    // The async operation.
+    temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
+}
+
+
+message DeleteExportSinkRequest {
+    // The namespace to which the sink belongs.
+    string namespace = 1;
+    // The name of the sink to delete.
+    string sink_name = 2;
```
anekkanti, line 724:
```suggestion
    string name = 2;
```

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -652,3 +653,95 @@ message GetUsageResponse {
...
+    repeated temporal.api.cloud.sink.v1.ExportSink sinks = 1;
+    // The next page token, set if there is another page.
+    string next_page_token = 2;
+}
+
+message UpdateExportSinkRequest {
+    // The namespace to which the sink belongs.
+    string namespace = 1;
+    // The name of the sink to update.
+    string sink_name = 2;
```
anekkanti, line 704:
```suggestion
    string name = 2;
```

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -305,4 +305,50 @@ service CloudService {
...
+        option (google.api.http) = {
+            post: "/cloud/namespaces/{namespace}/sinks",
+            body: "*"
+        };
+    }
+
+    // Get an export sink 
+    rpc GetExportSink(GetExportSinkRequest) returns (GetExportSinkResponse) {
+    option (google.api.http) = {
+            get: "/cloud/namespaces/{namespace}/sinks/{sink_name}"
```
anekkanti, line 320:
```suggestion
            get: "/cloud/namespaces/{namespace}/sinks/{name}"
```

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -305,4 +305,50 @@ service CloudService {
...
+    rpc GetExportSinks(GetExportSinksRequest) returns (GetExportSinksResponse) {
+        option (google.api.http) = {
+            get: "/cloud/namespaces/{namespace}/sinks"
+        };
+    }
+
+    // Update an export sink
+    rpc UpdateExportSink(UpdateExportSinkRequest) returns (UpdateExportSinkResponse) {
+    option (google.api.http) = {
+            post: "/cloud/namespaces/{namespace}/sinks/{sink_name}",
```
anekkanti, line 334:
```suggestion
            post: "/cloud/namespaces/{namespace}/sinks/{name}",
```

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -305,4 +305,50 @@ service CloudService {
...
+            body: "*"
+        };
+    }
+
+    // Delete an export sink
+    rpc DeleteExportSink(DeleteExportSinkRequest) returns (DeleteExportSinkResponse) {
+    option (google.api.http) = {
+            delete: "/cloud/namespaces/{namespace}/sinks/{sink_name}"
```
anekkanti, line 342:
```suggestion
            delete: "/cloud/namespaces/{namespace}/sinks/{name}"
```

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -652,3 +653,95 @@ message GetUsageResponse {
...
+message CreateExportSinkResponse {
+    // The async operation.
+    temporal.api.cloud.operation.v1.AsyncOperation async_operation = 2;
+}
+
+message GetExportSinkRequest {
+    // The namespace to which the sink belongs.
+    string namespace = 1;
+    // The name of the sink to retrieve.
+    string sink_name = 2;
```
anekkanti, line 675:
```suggestion
    string name = 2;
```

temporal/api/cloud/cloudservice/v1/service.proto
atihkin: Curious about the choice of` /sinks/ `. Do we expect more than 1 sink? I assumed we could only have 1.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -652,3 +653,95 @@ message GetUsageResponse {
...
+}
+
+message GetExportSinksResponse {
+    // The list of export sinks retrieved.
+    repeated temporal.api.cloud.sink.v1.ExportSink sinks = 1;
+    // The next page token, set if there is another page.
+    string next_page_token = 2;
+}
+
+message UpdateExportSinkRequest {
```
atihkin, line 700: Just confirming we can handle the MRN case correctly.

alice-yin: Yes. We could.

temporal/api/cloud/cloudservice/v1/service.proto
robzienert: Agree, while reading this PR I was wondering if there was a scope change that multiple sinks would be supported. 

Unless there's already an existing product requirement to support multiple sinks, I'd prefer to have a simpler API versus one that is future-proofing for a growth area that may never come. If a customer needs to sink multiple places, I think it's a reasonable expectation that they handle the fan-out themselves from a single source.

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -652,3 +653,95 @@ message GetUsageResponse {
...
+message CreateExportSinkRequest {
+    // The namespace under which the sink is configured.
+    string namespace = 1;
+    // The specification for the export sink.
+    temporal.api.cloud.sink.v1.ExportSinkSpec spec = 2;
+    // Optional. The ID to use for this async operation.
+    string async_operation_id = 3;
+}
+
+message CreateExportSinkResponse {
```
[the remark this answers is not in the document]
alice-yin: So the name is user given. It's unique for each namespace but not globally. Thus, I'm not sure it's a good value to return. What do you think?

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -652,3 +653,95 @@ message GetUsageResponse {
...
+    // The ID to use for this async operation - optional.
+    string async_operation_id = 4;
+}
+
+message DeleteExportSinkResponse {
+    // The async operation.
+    temporal.api.cloud.operation.v1.AsyncOperation async_operation = 1;
+}
+
+message ValidateExportSinkRequest {
```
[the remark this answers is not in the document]
alice-yin: User is expected to have their IAM role pre-created, and they send Validate sink request. One our side, we'll write an empty file to user's sink. If we failed, we will return an error. But the error is directly returned from API. Actually the response is not even an async operation. It should be empty. 

And Yes. Nothing change on the backend during this process. We're not going to save the sink information.

temporal/api/cloud/cloudservice/v1/service.proto
alice-yin: So user will be allowed to set up 2 sinks for multi region namespace, still one sink per region, but 2 sinks in total.
 
It's under a feature flag now. Not open to customer yet but backend support has completed. Would like to hear more feedback on this.

temporal/api/cloud/sink/v1/message.proto
```diff
@@ -0,0 +1,109 @@
...
+
+message ExportSinkSpec {
+    // The name of the export sink.
+    string name = 1;
+
+    // A flag indicating whether the export sink is enabled or not.
+    bool enabled = 2;
+
+    // The receiving sink type for export.
+    ExportDestinationType destination_type = 3;
```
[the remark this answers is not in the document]
alice-yin: Makes sense to me. Wondering how to change existing backend for it?

temporal/api/cloud/sink/v1/message.proto
alice-yin: Another thing is sink has its only internal storage. So it's not depends on namespace table. It's a sub resource of namespace. Thus, I'm hesitated to merge to namespace. 

Do you think we could move the sink folder under namespace folder?

temporal/api/cloud/cloudservice/v1/service.proto
atihkin: Ah that's a good point! In that case, `sinks` makes sense. Can we document in the API somewhere that by default export has only 1 sink? To minimize confusion

temporal/api/cloud/sink/v1/message.proto
cretz, line 1: This seems more like `temporal.api.cloud.export.v1` or `temporal.api.cloud.exportsink.v1` package rather than a general-purpose `sink` package.

temporal/api/cloud/sink/v1/message.proto
```diff
@@ -0,0 +1,91 @@
...
+    
+    // The S3 configuration details when destination_type is S3.
+    S3Spec s3 = 3;
+
+    // This is a feature under development. We will allow GCS sink support for GCP Namespaces.
+    // The GCS configuration details when destination_type is GCS. 
+    GCSSpec gcs = 4;
+}
+
+
```
cretz, line 31: Inconsistently adding two blank lines sometimes and a single blank line others

cretz (commented):
Mostly LGTM. I do think it's clearer in an "export" area instead of assuming some of those top-level S3/GCP models could be used for some future general purpose "sink".

temporal/api/cloud/sink/v1/message.proto
anekkanti: You could move it to a sub folder under the namespace folder. Just feels like it will be really thin for it to be its own package. Also we will be introducing a new pattern where we have sub packages (all our current packages are at root).

But we clearly need to move the   ExportSink and ExportSinkSpec to the namespace package (or namespace subpackage).
Just to make sure we are on the same page the `S3Spec` and `GCSSpec` will remain here in the sink package, correct?

alice-yin: The only thing is this change has already landed in saas-proto. Can we have a sync on this and see how strong you feel this needs to change?

temporal/api/cloud/sink/v1/message.proto
alice-yin: Also I think it's useful to keep a general sink package. As some of the sink spec could be used by other sink in the future. Currently, we also have audit log. I think audit log has pubsub and kinesis. I expect those specs will need to be in this sink package.

cretz: Yeah, I caught the HTTP URL being too generally named as "sinks" there and forgot to follow up to catch the gRPC package name having the same problem. I think it'd be ideal if it were in a place specific to export, but technically up to y'all.

cretz (commented):
Approving, though would prefer the models were in an export-specific place, but I missed it on an internal review.

cretz (approved):

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -211,3 +212,51 @@ message NamespaceRegionStatus {
...
+    // An error message describing any issues with the export sink, if applicable.
+    string error_message = 6;
+
+    // The timestamp of the latest successful data export.
+    google.protobuf.Timestamp latest_data_export_time = 7;
+
+    // The timestamp of the last health check performed on the export sink.
+    google.protobuf.Timestamp last_health_check_time = 8;
+
+    enum ExportSinkHealth {
```
anekkanti, line 256: Can we just call it `Health` as its nested under `ExportSink`?

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -211,3 +212,51 @@ message NamespaceRegionStatus {
         STATE_FAILED = 5;       // The region failed to be added/removed, check failure_reason in the last async_operation status for more details.
     }
 }
+
+message ExportSinkSpec {
+    // The unique name of the export sink.
+    string name = 1;
```
anekkanti, line 218: Can we add a comment about the mutability of each field in the Spec?

temporal/api/cloud/cloudservice/v1/request_response.proto
```diff
@@ -681,3 +681,89 @@ message UpdateAccountResponse {
...
+}
+
+message GetNamespaceExportSinksResponse {
+    // The list of export sinks retrieved.
+    repeated temporal.api.cloud.namespace.v1.ExportSink sinks = 1;
+    // The next page token, set if there is another page.
+    string next_page_token = 2;
+}
+
+message UpdateNamespaceExportSinkRequest {
```
anekkanti, line 727: Can we add the `name` field to the update request? Similar to how we need it for the delete request and get request. 
Makes things consistent, I think.

alice-yin: For the update I think it's more preferred to get it directly from the spec

anekkanti (approved):

PR #54, opened 2024-12-19
add enable_delete_protection flag to NamespaceSpec

mattkim:
## What was changed

add enable_delete_protection flag to NamespaceSpec

## Why?

allows customers to set the delete protection flag on namespaces to prevent accidental deletion

## Checklist
1. How was this tested:

e2e tests

3. Any docs updates needed?

Need to update docs on creating and updating namespaces to include delete protection: https://docs.temporal.io/cloud/namespaces#create-a-namespace


anekkanti (approved):

VERSION
cretz, line 1: Is this version already rolled out and therefore using this version today works? I am just worried about people that use https://raw.githubusercontent.com/temporalio/api-cloud/refs/heads/main/VERSION directly.

Semi-related general question - is it always compatible even if not released to cloud yet? Meaning what if I put `2026-01-01-00`, would it work?

cretz, line 1: Should make it clear in the PR title this is changing in this PR IMO for those reading lists of PRs as release notes

cretz (approved):
Nothing blocking

VERSION
anekkanti: The changes for `2025-01-01` are rolled out to prod. Its safe for clients to pick the version from the VERSION file directly.

And no you cannot send `2026-01-01-00` as things stand right now.

 But a lot of this changing, i am moving all these version to semver that will be the same as the release versions.

cretz: 👍 So long as the version in `main` works as soon as it's merged, works for me

cretz: Bump here, can we change this PR's title to include that we are increasing the version?

PR #55, opened 2025-01-06
[Depend on Temporal API][Breaking] Change Nexus Endpoint description type

nikki-dag:


buf.yaml
```diff
@@ -1,9 +1,10 @@
 version: v1
 deps:
   - buf.build/googleapis/googleapis
+  - buf.build/temporalio/api:v1.43.0
 breaking:
   use:
-    - FILE
+    - WIRE
```
nikki-dag, line 7: Will revert back with a follow up PR

buf.yaml
```diff
@@ -1,9 +1,10 @@
 version: v1
 deps:
   - buf.build/googleapis/googleapis
+  - buf.build/temporalio/api:v1.43.0
```
cretz, line 4: May also want to change issue title also say "Depend on Temporal API" since that's arguably the bigger change in this PR.

So my primary concern here is that we have tooling in Go, Java, Rust, TypeScript (I think?), Python, .NET, and Ruby for generating cloud protos. It's possible they could need slight tweaks with the added dependency, though not huge since they surely have the dependency right there.

As a test to at least make sure we don't break the Go one before we merge this, can you manually update the `api-cloud` submodule in https://github.com/temporalio/api-go/tree/master/proto to point to this branch and confirm the `Makefile` still works to generate the cloud code (confirming that the `EndpointSpec.Description` does indeed reference the proper `Payload` structure)?

nikki-dag: That's a very good point! I will do try it today and let you know.

nikki-dag: I was able to successfully run `make install` and `make proto` and the checks are passing:
https://github.com/temporalio/api-go/pull/201

cretz: Great, thanks!

cretz (approved):

PR #56, opened 2025-01-09
Add Publish to Buf Github Action

nikki-dag:
https://github.com/temporalio/api/pull/508#discussion_r1907391956

Tested with my PR branch:
image


.github/workflows/push-to-buf.yml
```diff
@@ -0,0 +1,18 @@
+name: Push to Buf Registry
+
+on:
+  push:
+    branches:
+      - master
```
cretz, line 6: Even though this has no tags yet, we should consider still triggering on them for the one day it actually might

buf.yaml
```diff
@@ -1,4 +1,5 @@
 version: v1
+name: buf.build/temporalio/api-cloud
```
cretz, line 2: As part of internal discussions concerning the future of cloud repos, I have proposed that if/when we get to multiple cloud-specific repos, they would all be prefixed with `cloud-` which includes renaming this repository to `cloud-api` instead of `api-cloud`. I wonder if we should think about making the buf repo as `cloud-api` now or at least confirm that we don't think it'll be too harmful to users if/when we do it later (but before cloud API is considered stable).

nikki-dag: If we eventually rename this repo to `cloud-api,` it makes sense to me to name the buf registry `cloud-api` now. What do you think, @anekkanti?

.github/workflows/push-to-buf.yml
nikki-dag: Updated ✅  (Assuming our version tags will also start with `v`)

buf.yaml
anekkanti: The repo has been changed to `cloud-api`. @nikki-dag please go ahead and change the buf registry to `cloud-api` as well.

.github/workflows/push-to-buf.yml
anekkanti: Yes we have started cutting release tags now, lets make sure we push them to buf too.

anekkanti (approved):

PR #58, opened 2025-01-16
Add Proto for remove region and Disaster recovery spec

yux0:
## What was changed
Add Proto for remove region and Disaster recovery spec

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


anekkanti (approved):

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -128,6 +128,13 @@ service CloudService {
         };
     }
 
+    // Remove a region from a namespace
+    rpc RemoveNamespaceRegion (RemoveNamespaceRegionRequest) returns (RemoveNamespaceRegionResponse) {
+        option (google.api.http) = {
+            delete: "/cloud/namespaces/{namespace}/{region}",
```
cretz, line 134:
```suggestion
            delete: "/cloud/namespaces/{namespace}/regions/{region}",
```

What if my region was named `add-region`? We have not properly reserved all children of `/cloud/namespaces/{namespace}` as regions, so it's not like `POST /cloud/namespaces/{namespace}` adds a region. Also, `POST /cloud/namespaces/{namespace}/add-region` is probably not the best HTTP API for adding, I'd consider adding an additional binding (or even making the BC change to remove the other) to `POST /cloud/namespaces/{namespace}/regions`.

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -128,6 +128,13 @@ service CloudService {
         };
     }
 
+    // Remove a region from a namespace
+    rpc RemoveNamespaceRegion (RemoveNamespaceRegionRequest) returns (RemoveNamespaceRegionResponse) {
```
cretz, line 132: I assume listing namespace region(s) is done via some other call?

yux0: Yes. We have GetRegions and GetRegion APIs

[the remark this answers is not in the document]
yux0: Updated

temporal/api/cloud/cloudservice/v1/service.proto
yux0: all make sense

cretz (approved):

temporal/api/cloud/namespace/v1/message.proto
```diff
@@ -109,6 +114,10 @@ message NamespaceSpec {
     // Optional, default is unset.
     CodecServerSpec codec_server = 6;
 
+    // The disaster recovery configuration for the namespace.
+    // temporal:versioning:min_version=2025-01-01-00
```
yux0, line 118: @anekkanti do you know if I need to change this version to v0.4.0?

anekkanti: Yes you do need to change the version to `v0.4.0`
Thanks!

temporal/api/cloud/cloudservice/v1/service.proto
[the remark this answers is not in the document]
anekkanti: I feel `RemoveNamespaceRegion` works better here. Since it complements better with `AddNamespaceRegion`.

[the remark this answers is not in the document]
cretz: I didn't think about the difference between `Create` and `Add` here. I'm ok leaving it as `Remove` even if the HTTP verb will remain `DELETE` across both `Delete` and `Remove` prefixed RPCs.

PR #59, opened 2025-01-21
Merge api versions with semver based release versions.

anekkanti:
## What was changed
This change will deprecate the date based api versions. And merge the api version with the release versions.

## Why?
To reduce cognitive overhead of dealing with 2 kind of versions namely api version and release versions on the clients.


anekkanti:
Note that the required back end changes required to translate the existing date based api-versions to the new semver based versioning for backward compatibility is already deployed.

captainbeardo (approved):

cretz (approved):

PR #108, opened 2026-03-24
release replicaspec into namespace spec

yux0:
## What was changed
add replicaspec into namespace spec

## Why?
add replicaspec into namespace spec

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


captainbeardo (approved):

temporal/api/cloud/cloudservice/v1/service.proto
```diff
@@ -285,7 +285,9 @@ service CloudService {
     }
 
     // Add a new region to a namespace
+    // Deprecated: Use the UpdateNamespace() to add new replica in the namespace spec instead.
```
laurakwhit, line 288: @yux0 saw this was deprecated, but when I try to add replicas to the namespace spec in a `CloudService.UpdateNamespace` request  I see this error ⬇️ 

```
failed to version request: rpc error: code = InvalidArgument desc = field 'Replicas' is not supported in version v0.14.0
```

Is there a version were we will need to have this updated in/by?

PR #128, opened 2026-08-10
Say "user" instead of "customer" in API docs; clarify Connectivity Rule fields

lukeknep:
These comments are published as the Cloud Ops API reference and as doc comments in every generated SDK binding, where "customer" is the wrong register for the person reading them.

Two corrections are not just wording:

- `RenameCustomSearchAttribute` said "customer search attribute", which was a typo for "custom" — the RPC cannot rename built-in attributes, and its own openapiv2 block already said "custom".
- `azure_pe_resource_id` is rejected when empty, so it is now documented as required for Azure private rules.

The Connectivity Rule descriptions were fixed (some were truncated before). The fix was to introduce a short title. See the screenshot for the thing being fixed. Context: `protoc-gen-openapiv2` routes a comment's first paragraph to the OpenAPI `title` and the remainder to `description`, so fields that ran on without one were rendering truncated in the API reference.
Screenshot 2026-08-10 at 2 39 10 PM

## AI Verification
Confirmed the diff is comment-only with no field, type, or number changes, and checked the Azure required-ness claim against `saas-control-plane`'s provider validation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


temporal/api/cloud/connectivityrule/v1/message.proto
```diff
@@ -27,53 +30,66 @@ message ConnectivityRule {
...
 
-// A private connectivity rule allows connections from a specific private vpc only.
+// A private Connectivity Rule allows access from a specific PrivateLink or PSC.
 message PrivateConnectivityRule {
-  // Connection id provided by user to enforce the private connectivity.
-  // For AWS: VPC endpoint ID (e.g. "vpce-0123456789abcdef0").
-  // For GCP: PSC connection ID (numeric string).
-  // For Azure: not set by the customer; populated internally with the PPv2 LinkID
-  // returned by the infra plane during Private Endpoint connection approval.
+  // Resource ID from AWS or GCP
```
jeri-temporal, line 67: The gRPC docs don't treat the first line as a title, so a missing period here generates a run-on sentence: 

> Region of the Connectivity Rule The region of the connectivity rule...

temporal/api/cloud/connectivityrule/v1/message.proto
```diff
@@ -27,53 +30,66 @@ message ConnectivityRule {
...
+  // The region of the connectivity rule.
+  // Must be the same region as the Namespace or its replica.
+  // Required for all cloud providers.
+  // Examples: "aws-us-west-2", "gcp-us-central1", "azure-eastus"
   string region = 3;
 
   reserved 4;
 
-  // azure_pe_resource_id is the ARM resource ID of the customer's Azure Private Endpoint.
+  // Resource ID from Azure
```
jeri-temporal, line 90: Missing periods here as well

temporal/api/cloud/connectivityrule/v1/message.proto
```diff
@@ -27,53 +30,66 @@ message ConnectivityRule {
...
-  // Connection id provided by user to enforce the private connectivity.
-  // For AWS: VPC endpoint ID (e.g. "vpce-0123456789abcdef0").
-  // For GCP: PSC connection ID (numeric string).
-  // For Azure: not set by the customer; populated internally with the PPv2 LinkID
-  // returned by the infra plane during Private Endpoint connection approval.
+  // Resource ID from AWS or GCP
+  //
+  // For AWS: the user's VPC endpoint ID (e.g. "vpce-0123456789abcdef0").
+  // For GCP: the user's PSC connection ID (numeric string).
+  // For Azure: users do not use this. Temporal Cloud will set a value after the Private Endpoint is Approved, for internal Temporal Cloud use, only.
```
jeri-temporal, line 71: IMO, `users do not use this... for internal Temporal Cloud use, only` is a jarring thing to see in public API docs. Maybe something like this? 

For Azure: Connection ID is automatically set when a connection to the specified Private Endpoint is Approved, and is not user-settable.

PR #129, opened 2026-08-25
Add authentication settings to account

asmclean:
## What was changed
Bump to `v0.20.0` and add `AuthenticationSettings` to `AccountSpec`.

## Why?
To include support for strict session lifetimes for Cloud accounts.

## Checklist

1. Closes CLDIAM-1024

2. How was this tested:

3. Any docs updates needed?


temporal/api/cloud/account/v1/message.proto
```diff
@@ -19,10 +19,18 @@ message MetricsSpec {
     bytes accepted_client_ca = 2;
 }
 
+message AuthenticationSettings {
+    // Restricts session lifetimes for users in this account.
```
jeri-temporal: Does this apply to all sessions, or just in-browser?

asmclean: @jeri-temporal, this will be just for the web UI as of now.

jeri-temporal: It might be worth clarifying that in the comment, since that'll make its way into the docs

jeri-temporal (approved):

